PR #3, opened 2017-08-02
Tt 876 log useful server urls

IreneLau-GDS:


IreneLau-GDS:
Yeah I did start off with trying to do that but it started to look ugly so it'd be good to pair up when you get a few minutes. I do want to make the ServerDetail object immutable!

IreneLau-GDS:
Have now made the suggested changes during "in between mobbing" time.

PR #15, opened 2017-08-23
TT-999 More logging

RachaelBooth:
Authors: @rachaelbooth


RachaelBooth:
@richardTowers It's only me who has looked at this, so it wants a proper review/think about.

I've checked, and I can't see anywhere that we might be logging stuff that we shouldn't be.
Now logs success or failure (with reason) of generate-request and translate-response calls. I'm not convinced by the logging in the exception mappers, in that they don't tell you where it came from, but it is better than nothing and I can't see a way to fix that (since the exception is thrown before you reach the body of the function).
(N.B. it should, from defaults and what's already been done, also log uncaught exceptions with stack traces, and the startup info, and things if healthchecks fail.)

As we deploy things we may find that more logging is useful, but I don't want to add in loads of clutter if no one thinks we need it (Andy agreed with me on that one).

richardTowers:
This looks good to me. Let's go for it.

PR #31, opened 2017-09-25
TT-1122 Added validation to ensure recipient matches the assertion consumer URI

SKeerthana:


src/main/java/uk/gov/ida/verifyserviceprovider/validators/SubjectValidator.java
```diff
@@ -52,6 +54,15 @@ public void validate(Subject subject, String expectedInResponseTo) {
             throw new SamlResponseValidationException(String.format("'InResponseTo' must match requestId. Expected %s but was %s", expectedInResponseTo, actualInResponseTo));
         }
 
+        String recipient = subjectConfirmationData.getRecipient();
+        if (recipient == null) {
+            throw new SamlResponseValidationException("Subject confirmation data must contain 'Recipient'.");
+        }
+
+        if (!assertionConsumerServiceUri.equals(recipient)) {
+            throw new SamlResponseValidationException(String.format("'Recipient' must match entity id. Expected %s but was %s", assertionConsumerServiceUri, recipient));
```
richardTowers: Error message should say "'Recipient' must match assertionConsumerServiceUri" not "entity id" right?

configuration/verify-service-provider.yml
```diff
@@ -25,6 +25,7 @@ clockSkew: ${CLOCK_SKEW:-PT5s}
 
 # Entity ID that uniquely identifies your service
 serviceEntityId: ${SERVICE_ENTITY_ID:-http://verify-service-provider}
+assertionConsumerServiceUri: ${ASSERTION_CONSUMER_SERVICE_URI:-http://localhost:3200/verify/response}
```
richardTowers: Can we put a comment above this one explaining what it is (like the others)? Something like:

"The URI that Verify should send SAML responses to"

SKeerthana: Yes. We also want to discuss about the sensible default value.

src/main/java/uk/gov/ida/verifyserviceprovider/validators/SubjectValidator.java
SKeerthana: Done

PR #34, opened 2017-09-27
TT-1260 Added an endpoint to read version number from manifest file 

SKeerthana:


src/test/java/unit/uk/gov/ida/verifyserviceprovider/resources/VersionNumberResourceTest.java
```diff
@@ -0,0 +1,50 @@
...
+        String versionNumber = "1.2.0";
+        when(attributes.getValue("Version")).thenReturn(versionNumber);
+        when(manifestReader.getManifest()).thenReturn(attributes);
+
+        Response response = resources.target("/version-number").request().get();
+
+        verify(manifestReader, times(1)).getManifest();
+        assertThat(response.getStatus()).isEqualTo(Response.Status.OK.getStatusCode());
+        assertThat(response.readEntity(String.class)).isEqualTo(versionNumber);
+    }
```
willp-bl: it's worth adding in a test if the manifest reader fails, too - some services have done _interesting_ things with msa jars in the past

SKeerthana: Done

PR #35, opened 2017-09-28
TT-1227: Parameterize entity id

RachaelBooth:
Many commits, a lot of them reverts. The git history is...interesting...


configuration/verify-service-provider.yml
```diff
@@ -24,10 +24,9 @@ logging:
 clockSkew: ${CLOCK_SKEW:-PT5s}
 
 # Entity ID that uniquely identifies your service
-serviceEntityId: ${SERVICE_ENTITY_ID:-http://verify-service-provider}
-
-# Location where Verify Hub would send the assertions to
-assertionConsumerServiceUri: ${ASSERTION_CONSUMER_SERVICE_URI:-http://localhost:3200/verify/response}
+serviceEntityId: [${SERVICE_ENTITY_ID:-"http://verify-service-provider"}]
+# To use the Verify Service Provider with multiple services, add their entity IDs here, e.g.
+# ["service-one-entity-id", "service-two-entity-id"]
```
RachaelBooth: I am slightly torn on leaving serviceEntityId singular, but it probably does make more sense for the mainline case of a single service.

src/main/java/uk/gov/ida/verifyserviceprovider/utils/ServiceEntityIdHelper.java
```diff
@@ -0,0 +1,58 @@
+package uk.gov.ida.verifyserviceprovider.utils;
```
RachaelBooth: I'm not convinced at all that this is a Helper, or that it should be in utils. (It morphed a bit after first writing).

Opinions on renaming/moving this class are particularly sought.

RachaelBooth: As are ones on rewording log messages. (Especially on rewording them in a way which would make this able to have one function, not two)

README.md
```diff
@@ -27,7 +27,7 @@ The following Environment Variables can be defined:
 ```
 #!/usr/bin/env sh
 
-export SERVICE_ENTITY_ID=... # The entity id of the service using Verify Service Provider
+export SERVICE_ENTITY_ID=... # The entity id of the service using Verify Service Provider. This will need to be in quotes.
```
andy-paine: Not necessarily clear from here that the _list_ needs to be in single quotes AND that they should really be double quote escaping each value (bash best practice). Nor is it particularly obvious that it is just a space separated list (although that is a fairly bash-y thing to do). Is there a reason we went for space separated as opposed to a JSON ["x", "y"] list?

architecture-decisions/adr/0022-we-will-not-validate-the-recipient.md
```diff
@@ -0,0 +1,23 @@
...
+
+In usual usages of SAML, the recipient should be validated, as should the audience restriction (if present), and these might be different and have slightly different meanings. However, since we control the production of this SAML (via the MSA), we know that the audience restriction will always be present and set to the service entity id, so validating this will mitigate any potential issues from not validating the recipient.
+
+## Decision
+
+The verify service provider will not validate the assertion consumer service url.
+
+## Consequences
+
+We will have to revert some commits already merged.
```
andy-paine: Probably worth mentioning that another consequence is that we are technically not performing a check that the SAML spec says we should

configuration/verify-service-provider.yml
andy-paine: Could we have `serviceEntityId` and `serviceEntityIds` and error if both are present?

src/main/java/uk/gov/ida/verifyserviceprovider/resources/TranslateSamlResponseResource.java
```diff
@@ -27,27 +30,33 @@
...
             TranslatedResponseBody translatedResponseBody = responseService.convertTranslatedResponseBody(
                 translateSamlResponseBody.getSamlResponse(),
                 translateSamlResponseBody.getRequestId(),
-                translateSamlResponseBody.getLevelOfAssurance()
+                translateSamlResponseBody.getLevelOfAssurance(),
+                entityId
             );
 
-            LOG.info(String.format("Translated response for requestID: %s, Scenario: %s",
+            LOG.info(String.format("Translated response for entityId: %s, requestID: %s, got Scenario: %s",
```
andy-paine: Should be consistent about `Id` vs. `ID`

src/main/java/uk/gov/ida/verifyserviceprovider/utils/ServiceEntityIdHelper.java
andy-paine: Seems like this whole class is a bit confused? It is a utility to abstract out the case of single entityId vs list of entityIds, but it is also doing validation that the entityId is valid. I think with it's current name, it is not clear what it is doing.

I would prefer this to just be a validator then we just do `validate(entityId)` line then just do an unchecked access to entityId afterwards (as we know it is valid if it made it past the validate line)

At the very least we can commonise the else block and flip the condition so that the single entityId case (what we expect to be normal) comes first (small processing optimisation?). Maybe make the 2 dtos inherit an interface that says they have an entityId on them? Don't know how is best to handle error messages then but it would bring the logic in line.

src/main/java/uk/gov/ida/verifyserviceprovider/utils/ServiceEntityIdHelper.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.List;
+
+public class ServiceEntityIdHelper {
+    private final List<String> configuredEntityIds;
+    private final String defaultEntityId;
+    private static final Logger LOG = LoggerFactory.getLogger(ServiceEntityIdHelper.class);
+
+    public ServiceEntityIdHelper(List<String> configuredEntityIds) {
+        this.configuredEntityIds = configuredEntityIds;
+        this.defaultEntityId = configuredEntityIds.size() == 1 ? configuredEntityIds.get(0) : null;
```
andy-paine: You can use streams -> findFirst and do this as an Optional to be more Java 8-y

src/test/java/feature/uk/gov/ida/verifyserviceprovider/configuration/ApplicationConfigurationFeatureTests.java
```diff
@@ -74,8 +75,42 @@ public void applicationShouldStartUp() throws Exception {
...
+            put("SAML_PRIMARY_ENCRYPTION_KEY", TEST_RP_PRIVATE_ENCRYPTION_KEY);
+            put("SAML_SECONDARY_ENCRYPTION_KEY", TEST_RP_PRIVATE_ENCRYPTION_KEY);
+            put("CLOCK_SKEW", "PT5s");
+        }});
+
+        application.getTestSupport().before();
 
+        VerifyServiceProviderConfiguration configuration = application.getConfiguration();
+
+        assertThat(application.getLocalPort()).isEqualTo(50555);
```
andy-paine: Can we pull this out into a constant field

src/main/java/uk/gov/ida/verifyserviceprovider/utils/ServiceEntityIdHelper.java
RachaelBooth: Error messages was my thinking - otherwise I would pass in the potentially null entity Id to a single function.

README.md
RachaelBooth: It shouldn't be space separated?

src/main/java/uk/gov/ida/verifyserviceprovider/utils/ServiceEntityIdHelper.java
```diff
@@ -0,0 +1,58 @@
...
+            LOG.info(String.format("Received request to generate authn request using default entityId"));
+            if (defaultEntityId != null) {
+                return defaultEntityId;
+            } else {
+                throw new InvalidEntityIdException("No entityId was provided, and there are several in config");
+            }
+        }
+    }
+
+    public String getEntityId(TranslateSamlResponseBody translateSamlResponseBody) {
```
SKeerthana: Except for the log message both the functions look the same. Can we extract the common functionality into a method?

RachaelBooth: They are, but the logging is annoyingly nested within the logic. We considered a refactor, but thought having to pass in a string to be used in the logging was a bit too dirty - hoping to find a rewording of the logging which makes that unnecessary.

src/test/java/feature/uk/gov/ida/verifyserviceprovider/configuration/ApplicationConfigurationFeatureTests.java
RachaelBooth: I'm actually going to say no to that - we could, but in a test I feel like it's clearer to show the exact thing we expect it to match, given that it's set as a string earlier in the test.

configuration/verify-service-provider.yml
RachaelBooth: This would be great. It turns out this is more complicated than I'm going to get done today, so I'm pushing an option which renames this to serviceEntityIds and just gets them to define a list, which may only have one element (because, on discussion, we thought that made more sense).

Should probably discuss having both - it would be nicer, but is probably more complicated config for users, even in the mainline case, and definitely more complicated code (it will be a bit messier in places, and we'll need to define our own validation thing sensibly which seems to be possible but more complicated than it originally sounded).

willp-bl: A list that generally has only one entry sounds ok to me, especially if it makes it easier to explain both use cases

RachaelBooth:
This does some refactoring, based on the above, although not all (see comments). 

It still has one config option for serviceEntityIds, which now takes an array. This seems reasonable - it would be possible to take either serviceEntityId or serviceEntityIds, but involve a lot of effort and make various bits of code more complicated. I'm happy if someone really wants that, but pragmatically I think the current state is easy to explain to anyone wishing to use it in either setup.

ServiceEntityIdHelper has been refactored a bit, but if anyone can think of a sensible rename please go ahead.

(I'm now away until Tuesday, so do merge or make changes and then merge without me...)

src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/AuthnRequestAcceptanceTest.java
```diff
@@ -27,38 +32,70 @@
...
-    public static DropwizardAppRule<VerifyServiceProviderConfiguration> application = new DropwizardAppRule<>(
+    public static final DropwizardTestSupport<VerifyServiceProviderConfiguration> singleTenantApplication = new DropwizardTestSupport<>(
         VerifyServiceProviderApplication.class,
         resourceFilePath("verify-service-provider.yml"),
         ConfigOverride.config("server.connector.port", String.valueOf(0)),
         ConfigOverride.config("logging.loggers.uk\\.gov", "DEBUG"),
         ConfigOverride.config("samlSigningKey", TEST_RP_PRIVATE_SIGNING_KEY),
-        ConfigOverride.config("hubSsoLocation", String.format("%s/%s", COMPLIANCE_TOOL_HOST, "SAML2/SSO"))
+        ConfigOverride.config("hubSsoLocation", String.format("%s/%s", COMPLIANCE_TOOL_HOST, "SAML2/SSO")),
+        ConfigOverride.config("serviceEntityIds", String.format("%s", SINGLE_ENTITY_ID))
```
richardTowers: Useless use of `String.format()`?

src/main/java/uk/gov/ida/verifyserviceprovider/resources/GenerateAuthnRequestResource.java
```diff
@@ -24,23 +25,26 @@
 
     private final URI ssoLocation;
     private final AuthnRequestFactory authnRequestFactory;
+    private final EntityIdService entityIdHelper;
```
richardTowers: Instance name is `...Helper` but class name is `...Service` - don't mind which we go for, but shouldn't they be consistent?

PR #45, opened 2017-10-09
Tt 1235 extensions

georgievh:


src/main/java/uk/gov/ida/verifyserviceprovider/VerifyServiceProviderApplication.java
```diff
@@ -67,12 +65,14 @@ public String getName() {
...
+        ManifestReader manifestReader = new ManifestReader();
 
         environment.jersey().register(new JerseyViolationExceptionMapper());
         environment.jersey().register(new JsonProcessingExceptionMapper());
         environment.jersey().register(new InvalidEntityIdExceptionMapper());
-        environment.jersey().register(new VersionNumberResource(new ManifestReader()));
-        environment.jersey().register(factory.getGenerateAuthnRequestResource());
+        environment.jersey().register(new VersionNumberResource(manifestReader));
+
+        environment.jersey().register(factory.getGenerateAuthnRequestResource(manifestReader));
```
richardTowers: Not really a problem, but to me the wiring up of the dependency of `VersionNumberResource` and `GenerateAuthnRequestResouce` on `ManifestReader` feels like something the factory should be doing, not the Application.

If `ManifestReader` is supposed to be a singleton then we can have a gross `synchronized` method in the factory, like we do with the metadata resolvers. Although looking at it it probably doesn't need to be.

src/main/java/uk/gov/ida/verifyserviceprovider/metadata/MetadataPublicKeyExtractor.java
```diff
@@ -0,0 +1,72 @@
...
+        this.metadataResolver = metadataResolver;
+        this.publicKeyFactory = publicKeyFactory;
+    }
+
+    public PublicKey getEncryptionPublicKey() {
+        try {
+            CriteriaSet criteria = new CriteriaSet(new EntityIdCriterion(entityId));
+            return Optional.ofNullable(metadataResolver.resolveSingle(criteria))
+                .map(this::getPublicKeys)
+                .orElseThrow(missingEntityIdException());
```
richardTowers: I think this would be slightly neater as `.orElseThrow(this::missingEntityIdException)` where `missingEntityIdException` produces a `RuntimeException` instead of a `Supplier<RuntimeException>`.

src/test/java/unit/uk/gov/ida/verifyserviceprovider/factories/saml/AuthnRequestFactoryTest.java
```diff
@@ -1,32 +1,68 @@
...
-    );
-
+    private static final ManifestReader manifestReader = mock(ManifestReader.class);
+    private static final EncrypterFactory encrypterFactory = mock(EncrypterFactory.class);
+    private static Encrypter encrypter;
+    private static Decrypter decrypter;
+    private static AuthnRequestFactory factory;
 
     @Before
     public void bootStrapOpenSaml() {
```
richardTowers: This method probably wants renaming now it's doing more than just bootstrapping. Happy with `before` or `setUp` - whatever the Cool Kids are calling their `@Before` methods these days.

src/main/java/uk/gov/ida/verifyserviceprovider/metadata/MetadataPublicKeyExtractor.java
```diff
@@ -0,0 +1,72 @@
...
+        this.entityId = entityId;
+        this.metadataResolver = metadataResolver;
+        this.publicKeyFactory = publicKeyFactory;
+    }
+
+    public PublicKey getEncryptionPublicKey() {
+        try {
+            CriteriaSet criteria = new CriteriaSet(new EntityIdCriterion(entityId));
+            return Optional.ofNullable(metadataResolver.resolveSingle(criteria))
+                .map(this::getPublicKeys)
```
richardTowers: Looking at this a bit more closely - `getPublicKeys` is going to throw a (not very helpful) exception if anything goes wrong, then we've got an `orElseThrow` that only handles the case where `resolveSingle` returns null, but throws a nice error.

Would it be better to remove the `orElseThrow` from `getPublicKeys`, change the `map` to a `flatMap` and have all the errors handled by `missingEntityIdException` ?

PR #46, opened 2017-10-09
Environment config option

RachaelBooth:
Reduce complexity of configuration by making it possible to set
hubSsoLocation, hub metadataUri, and other hub metadata configuration
options by setting a single value telling the VSP to run against
compliance tool, integration, or production.
This is a breaking config change.
Values can still be overridden from the yaml.


configuration/verify-service-provider.yml
```diff
@@ -29,14 +29,12 @@ serviceEntityIds: ${SERVICE_ENTITY_IDS:-["http://verify-service-provider"]}
...
-# Verify Service Provider consumes the metadata and
-# Hub Public certificates from the metadata to identify the hub
-verifyHubMetadata:
-  uri: ${HUB_METADATA_URL:-https://compliance-tool-reference.ida.digital.cabinet-office.gov.uk/SAML2/metadata/federation}
+# Verify Hub Environment. This tells the service provider where the Verify Hub
+# authentication flow begins and where to find the hub metadata the Verify Service
+# Provider consumes to identify the hub.
+# Valid values: COMPLIANCE_TOOL, INTEGRATION, PRODUCTION
+verifyHubConfiguration:
+  environment: ${HUB_ENVIRONMENT:-COMPLIANCE_TOOL}
```
RachaelBooth: It's not mentioned here (or the similar file in main/resources) that you can override the hubSsoLocation or metadata uri, expectedEntityId etc. We don't really expect RPs to do this, I think, but it should probably be documented somewhere for our own benefit - the main use will be in development. I am unsure where is best to put this though (it can be seen from the tests, but this feels insufficient).

src/test/java/unit/uk/gov/ida/verifyserviceprovider/configuration/MetadataConfigurationWithDefaultsTest.java
```diff
@@ -1,99 +0,0 @@
-package unit.uk.gov.ida.verifyserviceprovider.configuration;
```
RachaelBooth: These tests are now covered by the VerifyHubConfigurationFeatureTests or the MsaMetadataConfigurationTest.

configuration/verify-service-provider.yml
RachaelBooth: Note as well that though you can set the environment with an environment variable, you can't override the specifics with them (this is much as before, save that you had to set, and could therefore change, the sso location and metadata uri before). Overrides can be done using a yaml file, or using Java system properties if necessary (we do not recommend or document this, but it is a dropwizard default for that to be possible).

src/main/java/uk/gov/ida/verifyserviceprovider/configuration/HubMetadataConfiguration.java
```diff
@@ -150,26 +138,22 @@ private static String generateTrustStorePath(URI uri) {
                 trustStoreName = TEST_VERIFY_TRUSTSTORE_NAME;
                 break;
             default:
-                throw new RuntimeException("No trust store configured for Metadata URI " + uri);
+                throw new RuntimeException("No trust store configured for Hub Environment: " + hubEnvironment.name());
```
RachaelBooth: These exceptions are probably not strictly necessary - if the hubEnvironment was not recognised there would have been an exception thrown trying to parse it - but similar would have been the case before with the MetadataUri enum. I have left them, as I assume they were here in case we added more valid values to the enum without updating this (and this would be a much more useful error in that case). But I'll take them out if no one thinks there is a good reason for them to be here.

pen-test-manifest.yml
```diff
@@ -5,9 +5,8 @@ applications:
     env:
```
RachaelBooth: Do we still need this file?

src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/AuthnRequestAcceptanceTest.java
```diff
@@ -52,7 +52,7 @@
         ConfigOverride.config("server.connector.port", String.valueOf(0)),
         ConfigOverride.config("logging.loggers.uk\\.gov", "DEBUG"),
         ConfigOverride.config("samlSigningKey", TEST_RP_PRIVATE_SIGNING_KEY),
-        ConfigOverride.config("hubSsoLocation", String.format("%s/%s", COMPLIANCE_TOOL_HOST, "SAML2/SSO")),
+        ConfigOverride.config("verifyHubConfiguration.environment", "COMPLIANCE_TOOL"),
```
SKeerthana: Can we use the enum instead of the "COMPLIANCE_TOOL" string ?

configuration/verify-service-provider.yml
richardTowers: Yeah, I wonder if we should pull the `Advanced Configuration` section of the README out into a separate `.md` file. Then we could document all sorts of crazy configurations!

src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/AuthnRequestAcceptanceTest.java
RachaelBooth: We have to provide a string to the config override (config is read from strings after all). Could set it to HubEnvironment.COMPLIANCE_TOOL.toString() (or .name()), but it seems more normal to put it in as it would appear in a config file.

README.md
```diff
@@ -28,8 +28,7 @@ export PORT=... # The TCP port where the application will listen for HTTP traffi
 export LOG_LEVEL=... # The threshold level for logs to be written (e.g. DEBUG, INFO, WARN, or ERROR) (default: INFO)
 export MSA_ENTITY_ID=... # The SAML Entity Id that identifies the Relying Party's Matching Service Adapter
 export MSA_METADATA_URL=... # The URL to the Matching Service Adapter's SAML metadata.
-export HUB_METADATA_URL=... # The URL to the Verify Hub's SAML metadata. (default: https://www.signin.service.gov.uk/SAML2/metadata/federation)
-export HUB_SSO_LOCATION=... # The URL to send SAML AuthnRequests to the Verify Hub
+export HUB_ENVIRONMENT=... # The environment of the Verify Hub to run against - PRODUCTION, INTEGRATION, or COMPLIANCE_TOOL
```
richardTowers: I wonder if `VERIFY_ENVIRONMENT` might be a better name. Not sure people outside this team actually understand the term `HUB` the way we do.

pen-test-manifest.yml
richardTowers: Nah, we should kill it 🔥

src/main/java/uk/gov/ida/verifyserviceprovider/configuration/HubMetadataConfiguration.java
richardTowers: More errors is good errors in my opinion 👍

src/main/java/uk/gov/ida/verifyserviceprovider/configuration/HubMetadataConfiguration.java
```diff
@@ -36,26 +34,16 @@
...
-        this.trustStorePath = ofNullable(trustStorePath).orElseGet(() -> generateTrustStorePath(uri));
-        this.trustStorePassword = ofNullable(trustStorePassword).orElse(DEFAULT_TRUST_STORE_PASSWORD);
-        this.uri = uri;
-        this.minRefreshDelay = ofNullable(minRefreshDelay).orElse(60000L);
-        this.maxRefreshDelay = ofNullable(maxRefreshDelay).orElse(600000L);
-        this.expectedEntityId = generateExpectedEntityId(uri, expectedEntityId);
-        this.jerseyClientConfiguration = ofNullable(jerseyClientConfiguration).orElse(createClient());
-        this.jerseyClientName = ofNullable(jerseyClientName).orElse(HUB_JERSEY_CLIENT_NAME);
-        this.shouldLoadTrustStoreFromResources = (trustStorePath == null);
+    public HubMetadataConfiguration(HubEnvironment environment, MetadataConfigurationOverrides overrides) {
```
richardTowers: Do we have enough `@Jsonthings` for it to work out about the environment / overrides things? I thought we'd need the `@JsonProperty / @JsonCreator` annotations. I guess not though, otherwise it wouldn't work 🤔

RachaelBooth: This isn't created directly from JSON anymore, it's called by the VerifyHubConfiguration class after that has deserialised both the environment and the overrides. (That class also needs the environment to do a similar thing for the hubSsoLocation)

PR #50, opened 2017-10-10
Tt 1302 changes to vsp local config

sgreensmith:
the changes to the compliance tool init script are intended to make it easier to switch between testing/demo'ing VSP locally and on the PASS environment.

the script now looks for an argument representing the environment that the journey is being run from.  In addition, it expects a PID to be set as an environment variable.

There were also some corrections to the keys that were in the config file.


richardTowers:
I got a bit confused by the use of `PID` here - it's commonly used to mean "process id" in shell scripts... Also feels a little weird to expect the pid as an env var, but the environment as an argument - shouldn't they either both be one or t'other?

PR #51, opened 2017-10-16
TT-1400 change configuration to match MSA

sgreensmith:
- change the structure of the config.yml file
- rename some Environment variables
- change the reference to the MSA in config to be clearer
- change the implementation of the metadata configuration classes to
refer to ExpectedEntityId as matchingServiceAdapter.entityId and uri as
matchingServiceAdapter.metadataUri

Author: @sgreensmith


src/main/java/uk/gov/ida/verifyserviceprovider/configuration/MatchingServiceAdapterConfiguration.java
```diff
@@ -51,11 +51,21 @@ public String getTrustStorePassword() {
...
+    @Valid
+    public URI getMetadataUri() {
+        return metadataUri;
+    }
+
+    @NotNull
+    @Valid
+    public String getEntityId() {
+        return entityId;
     }
```
RachaelBooth: It looks like if we move the `@NotNull` and `@Valid` annotations to the fields rather than the getters, we can have this work (giving the more useful names in error messages) without needing the extra getters, which would be a bit nicer.

RachaelBooth:
Also, thinking about it, the README and the Advanced_Configuration doc should be updated to match the changes. And src/main/resources/verify-service-provider-env.yml should be updated (the more hidden configuration file...argh).

It probably makes sense to leave this to you when you're back in the office, Simon.

richardTowers:
Closing because our thinking has matured a bit since this work was played, and we've got more stories to look at making this and the verify-matching-service-adapter easier to configure.

PR #60, opened 2017-10-30
BAU: Update Release Notes

RachaelBooth:
To reduce release pain, we ought to be keeping this vaguely up to date as we add features (although it should always be checked over and maybe reworded when it comes to a release). It seems useful to be in the habit of at least adding a bullet point to remind us what features have been added since the last release.

I was going to add support address history in, then figured I may as well look back at the diff and see what else I can catch - I think this is it, but I'm not completely sure I've caught everything that will be user visible. It's based on the diff since the last tag, which also may well not actually match the last release - I'm not sure. This will all be smoother once we've done the first one using the newly discussed process.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -10,6 +10,21 @@ Release notes
 * Improve documentation
 * Improve healthcheck logging
 
+#### Configuration Changes:
+When using environment variables:
+* Replace SERVICE_ENTITY_ID with SERVICE_ENTITY_IDS, which is a JSON string array containing the entity id of the service (or services) using the Verify Service Provider (e.g. '["http://entity-id"]')
+* Remove HUB_METADATA_URL and HUB_SSO_LOCATION
+* Add VERIFY_ENVIRONMENT, specifying the environment of the Verify Hub to run against. Valid values are PRODUCTION, INTEGRATION, or COMPLIANCE_TOOL.
+
+When using a yaml file:
+* Replace serviceEntityId with serviceEntityIds, which is a list containing the entity id (or ids) as above
+* Remove hubSsoLocation and verifyHubMetadata
+* Add verifyHubConfiguration as below. This will contain an environment option specifying which hub environment to use, removing the need to specify the hubSsoLocation or metadata url.
+```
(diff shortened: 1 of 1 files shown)


RELEASE_NOTES.md
```diff
@@ -3,6 +3,28 @@ Release notes
 
 ### Next
 
+* Add support for the Address History user account creation attribute
+* Add ENVIRONMENT configuration option to replace hubSsoLocation and metadataUrl.
+* Send version number to hub
+* Support multitenancy
```
richardTowers: Can we say "Support multiple services connected to GOV.UK Verify" like we do in the README?

RELEASE_NOTES.md
```diff
@@ -3,6 +3,28 @@ Release notes
 
 ### Next
 
+* Add support for the Address History user account creation attribute
+* Add ENVIRONMENT configuration option to replace hubSsoLocation and metadataUrl.
```
richardTowers: Maybe a `(see Configuration Changes below)` would fit here.

RELEASE_NOTES.md
```diff
@@ -3,6 +3,28 @@ Release notes
...
+* Send version number to hub
+* Support multitenancy
+* Improve documentation
+* Improve healthcheck logging
+
+#### Configuration Changes:
+When using environment variables:
+* Replace SERVICE_ENTITY_ID with SERVICE_ENTITY_IDS, which is a JSON string array containing the entity id of the service (or services) using the Verify Service Provider (e.g. '["http://entity-id"]')
```
richardTowers: Should probably mention that these block capital things are environment variables?

RELEASE_NOTES.md
```diff
@@ -3,6 +3,28 @@ Release notes
...
+* Remove HUB_METADATA_URL and HUB_SSO_LOCATION
```
richardTowers: ... environment variables, use  `VERIFY_ENVIRONMENT` to specify your environment instead (see below)

richardTowers:
Aww man, I was too slow with my comments - not super important I guess. /shrug

PR #61, opened 2017-10-31
TT-1233: Refactored code to have unit tests testing the relevant unit…

georgievh:


src/main/java/uk/gov/ida/verifyserviceprovider/utils/DateTimeComparator.java
```diff
@@ -19,19 +19,11 @@ public boolean isBeforeFuzzy(DateTime source, DateTime target) {
         return source.isBefore(target.plus(clockSkew));
     }
 
-    public boolean isDefinitelyBeforeNow(DateTime dateTime) {
+    public boolean isBeforeNow(DateTime dateTime) {
```
RachaelBooth: I see I'm a bit late to comment, but I'm not convinced of this rename - just because I'd expect isBeforeFuzzy(someTime, now) to be the same as isBeforeNow(someTime), and they aren't (while it's more obvious with some kind of qualifier in the name of this method). Probably doesn't matter too much though.

PR #65, opened 2017-11-06
adr-0024: We will only release one configuration file

richardTowers:
Authors: @richardtowers


chrisholmes:
Sounds good to me. Happy to merge soon unless there's alternative opinions.

PR #67, opened 2017-11-07
TT-1400: Commonise trust store config between MSA and VSP

andy-paine:
Modified configuration classes to use new MetadataConfiguration
format from saml-metadata-bindings v58.
Documented change in RELEASE_NOTES.md
Changed some tests to work with slightly greedier NotNull
validation

Authors: @andy-paine


src/main/java/uk/gov/ida/verifyserviceprovider/configuration/HubMetadataConfiguration.java
```diff
@@ -1,147 +1,70 @@
...
+        this.trustStoreConfiguration = trustStoreConfiguration;
     }
 
-    @NotNull
-    @Valid
-    @Override
-    public String getTrustStorePath() {
-        return trustStorePath;
+    public HubMetadataConfiguration() {
+        this(null, null, null, null, null, null, null, null);
```
andy-paine: This is just to make the `ofNullable(hubMetadataConfiguration).orElse(new HubMetadataConfiguration())` read a bit nicer but it is pretty ugly overall. Maybe a factory method somewhere would be nicer?

PR #68, opened 2017-11-08
TT-1574 Use a single config file

RachaelBooth:
Remove verify-service-provider-env.yml - the main configuration file will manage
environment variable substitutions. This means we only have one config file
to keep up to date in the repo.
If service provider is started with no arguments, use the configuration file at
path_to_jar/../verify-service-provider.yml.

Authors: @rachaelbooth


src/main/resources/verify-service-provider-env.yml
```diff
@@ -1,54 +0,0 @@
-# This configuration will be used when no YAML file is specified.
```
RachaelBooth: These are all the same environment variables mentioned in configuration/verify-service-provider.yml

src/main/java/uk/gov/ida/verifyserviceprovider/VerifyServiceProviderApplication.java
```diff
@@ -24,19 +24,14 @@
...
-
     public VerifyServiceProviderApplication() {
-        this(true);
     }
 
     public static void main(String[] args) throws Exception {
         if (Arrays.asList(args).isEmpty()) {
-            new VerifyServiceProviderApplication(false).run("server", "verify-service-provider-env.yml");
+            String jarPath = VerifyServiceProviderApplication.class.getProtectionDomain().getCodeSource().getLocation().getPath();
+            String configFilePath = jarPath.replaceFirst("/lib/verify-service-provider-.*\\.jar", "/verify-service-provider.yml");
```
RachaelBooth: Any less horrendous suggestions for this are appreciated.

README.md
```diff
@@ -116,6 +125,8 @@ __Build a distribution__
 
 You can find the distribution zip at `build/distributions`.
 
+To run or debug Verify Service Provider from an IDE, you will need to either delegate the run action to gradle or set a run configuration which uses the arguments ```server configuration/verify-service-provider.yml``` to provide the configuration file.
```
richardTowers: Is this true? Is that because the file is in `configuration` when you run it from the IDE?

src/main/java/uk/gov/ida/verifyserviceprovider/VerifyServiceProviderApplication.java
RachaelBooth: .getParent doesn't appear to be a function...

RachaelBooth: Oh, I see, VMSTT goes via creating a File object - I thought that seemed even messier, although happy to swap for consistency.

PR #69, opened 2017-11-08
TT-1598: Add AuthnRequest and reformat example responses

J-Lambo:
TT-1598: Publish AuthnRequest and SAML response in full

This was originally published without the full examples. Adding these for more clarity. 

Previously paired with: @richardTowers @andy-paine


J-Lambo:
@sgreensmith I think you're right. I've removed as it's more useful for the user to open the accordion than to click to hyperlink. 

We can re-test this content. If users want to go to the passport-verify repo, we can add a link somewhere else. What do you think?

PR #74, opened 2017-11-13
Remove defaults

RachaelBooth:


RachaelBooth:
This still leaves the old default values in the repository, which is not perfect, but not in the config file that will be distributed. But there seem to be use cases for the old defaults (for running against comp tool) and for using a local.env file from verify-local-startup (for running all of hub etc. locally) (currently the stuff in verify-local-startup needed for that is on a branch). 

It does sadly increase the number of things that need keeping up to date =(.

src/test/java/feature/uk/gov/ida/verifyserviceprovider/configuration/HubMetadataFeatureTest.java
```diff
@@ -67,6 +72,15 @@ public void setUp() {
...
         );
 
+        environmentHelper.setEnv(new HashMap<String, String>() {{
+            put("VERIFY_ENVIRONMENT", "COMPLIANCE_TOOL");
+            put("MSA_METADATA_URL", "some-msa-metadata-url");
+            put("MSA_ENTITY_ID", "some-msa-entity-id");
+            put("SERVICE_ENTITY_IDS", "[\"http://some-service-entity-id\"]");
+            put("SAML_SIGNING_KEY", TEST_RP_PRIVATE_SIGNING_KEY);
+            put("SAML_PRIMARY_ENCRYPTION_KEY", TEST_RP_PRIVATE_ENCRYPTION_KEY);
+        }});
```
andy-paine: What is the reason for using environment variables here over config overrides? If it is to test that the env var substitutions work then that is fine but I think it is better to stay consistent otherwise.

local-running/local-config.yml
```diff
@@ -0,0 +1,45 @@
...
+
+serviceEntityIds: ${SERVICE_ENTITY_IDS:-["http://dev-rp.local/SAML2/MD"]}
+
+verifyHubConfiguration:
+  environment: ${VERIFY_ENVIRONMENT:-COMPLIANCE_TOOL}
+  ssoLocation: ${FRONTEND_URI}/SAML2/SSO
+  metadata:
+    trustStore:
+      path: ../verify-local-startup/data/pki/metadata.ts
+      password: marshmallow
```
andy-paine: Even better - you can make this:
```
trustStore:
  type: encoded
  store: ${METADATA_TRUSTSTORE}
  password: ${METADATA_TRUSTSTORE_PASSWORD}
```
which avoids the `../verify-local-startup` reference

local-running/local-config.yml
```diff
@@ -0,0 +1,45 @@
...
+  level: ${LOG_LEVEL:-INFO}
+  appenders:
+    - type: console
+    - type: file
+      currentLogFilename: logs/verify-service-provider.log
+      archivedLogFilenamePattern: logs/verify-service-provider.log.%d.gz
+
+clockSkew: ${CLOCK_SKEW:-PT5s}
+
+serviceEntityIds: ${SERVICE_ENTITY_IDS:-["http://dev-rp.local/SAML2/MD"]}
```
SKeerthana: Are we not removing defaults from this config ?

RachaelBooth: We could for consistency - it might mean making the verify-local-startup script more complicated, but probably not by much. But this is the new file specifically for starting locally (importantly, not a file which will be used by relying parties, and it shouldn't be in the distZip), so it doesn't run up against the rp-config-having-defaults-in issue.

local-running/local-config.yml
RachaelBooth: Is the metadata truststore password env variable actually being set? I can't see it.

local-running/local-config.yml
RachaelBooth: That said, this one is defined in verify-local-startup, so I'll take it out. The environment I think is the one which isn't (we're overriding most things from that in this config file anyway).

PR #80, opened 2018-01-17
BAU: Build on Jenkins2 with OpenJDK

roryirvine:
As we'll be using OpenJDK as the basis of our alphagov/java8 build
container on Jenkins2, we need to ensure that VSP will behave as
expected when built in this way.

Note that the build.jenkins in this commit does not contain tag or
publish stages.

Authors: @roryirvine

### DIFF
--- a/build.jenkins
+++ b/build.jenkins
@@ -1,0 +1,39 @@
+pipeline {
+
+    agent none
+
+    options {
+        disableConcurrentBuilds()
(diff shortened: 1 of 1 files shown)


roryirvine:
We've not seen any errors during the test and build phases, but have had some reports from @andy-paine that running this in the OpenJDK JRE can cause errors.

We should therefore continue to be careful when extending this build.jenkins Jenkinsfile to publish an artefact for use in production.

PR #83, opened 2018-03-05
refactor: use public key derivers from OpenSaml

chrisholmes:
We're implemented our own code for deriving a public key from a private
key, but OpenSAML already provides a nice one-liner for doing this.
Therefore, I've replaced our code and done a little tidy up.


src/test/java/unit/uk/gov/ida/verifyserviceprovider/factories/saml/AuthnRequestFactoryTest.java
```diff
@@ -11,7 +11,9 @@
 import org.opensaml.saml.saml2.encryption.Decrypter;
 import org.opensaml.saml.saml2.encryption.Encrypter;
 import org.opensaml.security.credential.BasicCredential;
+import org.opensaml.security.crypto.KeySupport;
```
richardTowers: Who knew! 👍 How on earth do you find all these random opensaml features?

PR #94, opened 2018-07-05
49: Upgrade Dropwizard to 1.3.5

adityapahuja:
Upgrade Dropwizard to 1.3.5 to fix various security issues.
Update Dropwizard links to use 1.3.5.
Update all test cases to create an instance of client and initialise it before class to avoid getting uninitialised environment.

Authors: @adityapahuja


adityapahuja:
@andy-paine : It was not a refactor since these tests failed after upgrading Dropwizard to 1.3.5. I had to fix these tests so that they passed.

IreneLau-GDS:
Looking at the changes, they are acceptable in a single commit because they might touch many files the changes are small. (Plus the tests have to pass...)

PR #98, opened 2018-08-22
CVQ2-140 Option to initialise compliance tool with a matching dataset

Rosa-Fox:
Co-authored-by: Alan Carter <alan.carter@digital.cabinet-office.gov.uk>

PR #99, opened 2018-08-24
CVQ2 141 specify pid in json

Rosa-Fox:
- DO NOT MERGE until [CVQ2 140](https://github.com/alphagov/verify-service-provider/pull/98) has been merged.

- Specify PID in JSON file for non-matching and as an argument for matching.

### DIFF
--- a/scripts/init-compliance-tool-for-environment.sh
+++ b/scripts/init-compliance-tool-for-environment.sh
@@ -2,18 +2,22 @@
 set -e
 
 if [ ! $# -eq 3 ] ; then
-	echo "this script requires the environment and a persistentId to be passed in as arguments, along with either the '--matching' flag or the name of a file containing matching dataset json. e.g."
-	echo "$ ./init-compliance-tool.sh <environment> <persistentId> --matching"
+	echo "For a matching journey this script must include the environment, the --matching flag and a PID"
+	echo "For example: $ ./init-compliance-tool-for-environment.sh <environment> --matching <persistentId>"
+	echo "For a non-matching journey this script must include the environment, the --nonmatching flag and the filepath to the matching dataset json"
+	echo "For example: $ ./init-compliance-tool-for-environment.sh <environment> --nonmatching filepath/to/matching-dataset-for-rp.json"
 	echo "please try again, currently allowed environments are 'local' and 'dev'"
 	exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


alan-gds:
I'd encourage reviewers to have a think about whether we've taken the right approach in terms of where the PID (and other config) gets specified.

We don't want to PID to be specified in two places at once - that would be asking for trouble - so we don't take it as a parameter for the non-matching journey, as it's already required in the matching dataset json file.  Of course, we could have done the opposite and removed it from the matching dataset json file.  I think what we've done is more intuitive, since the json file now contains everything that is going to be returned in the test case responses (and nothing else besides).

It's worth keeping in mind that this script is only for our own convenience and is not likely to be used by RPs (although I presume they can see it if they're interested as the repo is public).  I think the current split between stuff specified in the script and stuff specified in the json file should be the most convenient one, but YMMV.

scripts/init-compliance-tool-for-environment.sh
```diff
@@ -1,57 +1,88 @@
 #!/usr/bin/env bash
 set -e
 
-if [ ! $# -eq 2 ] ; then
-	echo "this script requires the environment and a persistentId to be passed in as arguments. e.g."
-	echo "$ ./init-compliance-tool.sh <environment> <persistentId>"
+if [ ! $# -eq 3 ] ; then
+	echo "For a matching journey this script must include the environment, the --matching flag and a PID"
+	echo "For example: $ ./init-compliance-tool-for-environment.sh <environment> --matching <persistentId>"
+	echo "For a matching journey this script must include the environment, the --nonmatching flag and the filepath to the matching dataset json"
```
RacingHippo: Shouldn't that say "For a non-matching journey"?

alan-gds: Good spot; now corrected.

karlbaker02:
I agree that it's probably more intuitive that everything is in one place, although a counter argument could be made that to test different PIDs, initialising this via a command might be a quicker thing to change than the matching JSON file (in addition to giving you a history of initialisations of the compliance tool for varying PIDs, as opposed to changing the PID in a file, which has the effect of hiding the initialisation configurations).

In reality I think the setup of the PID in the JSON file is fine and is unlikely to cause any problems. Pending any changes that need to be made to PR #98, this looks good.

PR #100, opened 2018-09-20
Cvq2-161 Independent matching

RacingHippo:
Updates to the API definition (Swagger) in order to support both the existing journeys and the new "independent matching" journey. Changes are described in RFC51 -https://github.com/alphagov/verify-architecture/pull/51


docs/api/verify-service-provider-api.swagger-v2.yml
```diff
@@ -0,0 +1,293 @@
...
+        levelOfAssurance:
+          $ref: '#/components/schemas/RequiredLevelOfAssurance'
+        entityId:
+          $ref: '#/components/schemas/EntityId'
+    TranslatedResponseBody:
+      type: object
+      required:
+        - scenario
+        - pid
+        - attributes
```
RacingHippo: Thinking about this... if the scenario is anything other than IDENTITY_VERIFIED, we won't have a PID or any attributes. I'll take 'em out.

chrisholmes:
I'm closing this in favour of #184

PR #101, opened 2018-09-21
Run with local compliance tool

alan-gds:


alan-gds:
The acceptance tests are failing on this branch, and I think the failures are legitimate; I can't think of a reasonable way to resolve the test failures while keeping our changes.

Rosa and I had a go at creating a new branch where we were more selective about which bits of this work we included - CVQ2-228-initialise-local-compliance-tool - but that hasn't worked out either.  Leaving out some of the changes resulted in the VSP not being able to find the hub metadata service.  I don't think this is easily solvable, and in my view it's time to give up.  We can keep this branch and the original run-locally one, and make sure changes on master are regularly rebased onto this one.  CVQ2-228 and its associated branch can go.

I'm closing this PR.

PR #103, opened 2018-10-22
BAU: Add more example SAML Assertions

andy-paine:
- We will need to support 4 different kinds of assertion eventually so
  add some examples to reference off when trying to process them
- Adds both types of IDP assertion, a country assertion and an MSA
  assertion (there is already one of these but nice to regenerate in
case it has changed)

Co-authored-by: Olakunle Jegede <olakunle.jegede@digital.cabinet-office.gov.uk>


andy-paine:
It seems nobody really wants this and the PR encourager is noisy

PR #106, opened 2018-10-31
KBddnTLg translate assertions in the vsp

jkunle:


src/main/java/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionService.java
```diff
@@ -107,14 +116,6 @@ public void validateIdpAssertion( Assertion assertion,
...
-        if (!assertion.getAuthnStatements().isEmpty()) {
-            return AssertionType.AUTHN_ASSERTION;
-        } else {
-            return AssertionType.MDS_ASSERTION;
-        }
-    }
-
     private TranslatedResponseBody translateAssertions( List<Assertion> assertions ) {
         return null;
     }
```
alan-gds: We don't need this anymore.

src/main/java/uk/gov/ida/verifyserviceprovider/services/AssertionDataTranslator.java
```diff
@@ -0,0 +1,12 @@
+package uk.gov.ida.verifyserviceprovider.services;
```
alan-gds: Is this file necessary?

I'm actually a fan of having interfaces for everything, but that's not a pattern we seem to be keen on in Verify, and this one isn't being referenced anywhere (yes, one class extends it, but there are no references to the interface as a type).

src/main/java/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionService.java
```diff
@@ -123,4 +124,18 @@ private TranslatedResponseBody translateAssertions( List<Assertion> assertions )
     public TranslatedResponseBody translateNonSuccessResponse( StatusCode statusCode ) {
         return null;
     }
```
alan-gds: What's the plan for non-success responses?  We should discuss this with Andy.

src/main/java/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionService.java
jkunle: Yes the above logic, is implemented by a slight variation of your AssertionClassifier class only difference is that it has static members. Or maybe I don't get what you mean

src/main/java/uk/gov/ida/verifyserviceprovider/services/AssertionDataTranslator.java
jkunle: Maybe it isn't needed

src/main/java/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionService.java
jkunle: Yes, I think we should or maybe we are focusing on the happy path for now

src/main/java/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionService.java
alan-gds: private TranslatedResponseBody translateAssertions( List<Assertion> assertions ) {
        return null;
    }

jkunle: Ok, i think i'll go with your recommendations, the private method was been used by the below, so for now I'll replace the call it it with return null;, instead of return translateAssertions(assertions)

  @Override
    public TranslatedResponseBody translateSuccessResponse( List<Assertion> assertions, String expectedInResponseTo, LevelOfAssurance expectedLevelOfAssurance, String entityId ) {
        validate(assertions, expectedInResponseTo, expectedLevelOfAssurance);
        return null;
    }

I'll also remove the interface **AssertionDataTranslator**

alan-gds: Oh, I see, I hadn't realised anything was calling it.  I think something has gone quite wrong here.  translateSuccessResponse() should not be returning null.  Nothing is calling the translate() method that I wrote.  I think we took quite different approaches on the two branches, and they have not been integrated correctly.

src/test/java/unit/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionServiceTest.java
```diff
@@ -202,6 +211,45 @@ public void shouldNotThrowExceptionsWhenAssertionsAreValid() {
         verify(hubSignatureValidator, times(2)).validate(any(), any());
     }
 
+    @Test
+    public void shouldExpectAListContainingAnAuthnAssertion() {
+        Assertion authnAssertion = anAuthnStatementAssertion(IdaAuthnContext.LEVEL_2_AUTHN_CTX, "requestId").buildUnencrypted();
+        List<Assertion> assertions = Arrays.asList(authnAssertion);
+
+        exception.expect(NullPointerException.class);
```
Rosa-Fox: Why does it expect a `NullPointerException.class`? and also why does the `nonMatchingAssertionService.translate(assertions);` happen afterwards?

src/main/java/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionService.java
alan-gds: Ola's explained this to me; I was confused because I'd misunderstood the scope of the story.  We're not going to return a response yet, so it's fine that we're currently returning null.

For the same reason, it's not time to write any acceptance tests yet.

src/test/java/unit/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionServiceTest.java
alan-gds: What this is supposed to mean is that a NullPointerException is expected to be thrown later in the test (i.e. when `nonMatchingAssertionService.translate(assertions)` is called).  That's meant to happen because an assertion is missing.  It *is* happening, but it looks to me like these tests are actually back to front: the one which is meant to test a missing MDSassertion instead tests a missing AuthnAssertion, and vice versa.

Part of the reason this is all so confusing is that I've been lazy and just let the code throw an ordinary NullPointerException; it would be clearer if a custom exception were thrown instead.

alan-gds:
I've just noticed one other issue - the test for AssertionClassifierService from my branch has not been included.

src/test/java/unit/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionServiceTest.java
jkunle: So essentially these test are passing as is because the translate(assertions) is expecting 2 assertions instead of one. Otherwise the assertions shouldn't be interchangeable.

jkunle:
AssertionClassifier has static members, Andy recommended this; but this means it becomes tricky to test. Do you have any experience test static methods

src/test/java/unit/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionServiceTest.java
alan-gds: Well, that's not quite true.  If you give it multiple MDSAssertions, it'll still NPE because it can't find an AuthnAssertion.  This probably needs to be made clearer in the test cases though.

src/test/java/unit/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionServiceTest.java
```diff
@@ -91,16 +91,17 @@ public void setUp() {
...
-                attributeStatementValidator
+                attributeStatementValidator,
+                authnContextFactory,
+                verifyMatchingDatasetUnmarshaller,
+                assertionClassifier
         );
         doNothing().when(subjectValidator).validate(any(), any());
         when(hubSignatureValidator.validate(any(), any())).thenReturn(mock(ValidatedAssertions.class));
-
+        when(assertionClassifier.classifyAssertion(any())).thenCallRealMethod();
```
andy-paine: Might as well just instantiate this as a real class. I don't think this hurts the tests too much.

src/test/java/unit/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionServiceTest.java
andy-paine: We should probably wrap this in a nicer exception so that it logs nicer or we use some static typing and make `validate` do all these checks. If `validate` returned a `ValidatedResponse` and `translate` took that validated response, we could rely on the assertions being correct (i.e. 1 MDS, 1 Authn) and use the type system to make sure nobody is calling `translate` on unvalidated assertions. This will be tracked as tech debt so no need to deal with this as part of this card.

src/test/java/unit/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionServiceTest.java
```diff
@@ -202,6 +203,47 @@ public void shouldNotThrowExceptionsWhenAssertionsAreValid() {
...
+
+    @Test
+    public void shouldCorrectlyExtractLevelOfAssurance() {
+        Assertion authnAssertion = anAuthnStatementAssertion(IdaAuthnContext.LEVEL_2_AUTHN_CTX, "requestId").buildUnencrypted();
+        Assertion mdsAssertion = aMatchingDatasetAssertion("requestId").buildUnencrypted();
+        List<Assertion> assertions = Arrays.asList(authnAssertion, mdsAssertion);
+
+        AssertionData assertionData = nonMatchingAssertionService.translate(assertions);
+
+        verify(authnContextFactory, times(1)).authnContextForLevelOfAssurance(IdaAuthnContext.LEVEL_2_AUTHN_CTX);
```
andy-paine: We are potentially missing a little bit of logic - we are checking if the authnContextFactory method was called, but we are not checking if the result was put into the `AssertionData` object. We could use `when(authnContextFactory).thenReturn(...` and then `assertThat(assertionData.getLevelOfAssurance()).isEqualTo(...` which would allow us to test what actually gets put in the object.

src/main/java/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionService.java
andy-paine: @andy-paine will raise a card to deal with this at some point

src/test/java/unit/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionServiceTest.java
jkunle: We'll deal with this when we have a card for it. Investigated wrapping this in a nicer exception though but turned out a little messy. So abandoned it.

src/test/java/unit/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionServiceTest.java
```diff
@@ -15,28 +15,31 @@
...
-import uk.gov.ida.verifyserviceprovider.validators.InstantValidator;
 import uk.gov.ida.verifyserviceprovider.validators.SubjectValidator;
 
+import java.util.Arrays;
 import java.util.List;
 
 import static java.util.Arrays.asList;
 import static java.util.Collections.emptyList;
-import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Java6Assertions.assertThat;
```
andy-paine: Probably want to keep using `org.assertj.core.api.Assertions.assertThat` rather than Java6

jkunle: Will do thanks

PR #108, opened 2018-11-07
aBHXEwu8: Added acceptance tests for non matching scenarios

AmritSidhu:
All tests pass with status code 200
Created builders for v2 nonmatching to test more efficiently
Added scenario ID's from compliance tool
Created a v2 dataset for nonmatching to test

Co-authored-by: Andy Paine <andy.paine@digital.cabinet-office.gov.uk>

### DIFF
--- a/src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/AuthnRequestAcceptanceTest.java
+++ b/src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/AuthnRequestAcceptanceTest.java
@@ -22,7 +22,7 @@ import static javax.ws.rs.core.Response.Status.OK;
 import static org.assertj.core.api.Assertions.assertThat;
 import static uk.gov.ida.saml.core.test.TestCertificateStrings.TEST_RP_PRIVATE_ENCRYPTION_KEY;
 import static uk.gov.ida.saml.core.test.TestCertificateStrings.TEST_RP_PRIVATE_SIGNING_KEY;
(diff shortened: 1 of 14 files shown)


alan-gds:
I don't understand why we now have two essentially identical VerifyServiceProviderAppRuleBuilder classes.  Surely the only areas where the tests are going to differ for V2 will be around communicating with the ComplianceToolService, which isn't affected the AppRuleBuilders.

src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/builders/V2MatchingDataset.java
```diff
@@ -0,0 +1,38 @@
+package uk.gov.ida.verifyserviceprovider.builders;
```
alan-gds: This class isn't a builder and doesn't call itself one, so it probably shouldn't be in this package.  I'll probably need to turn it into a builder in my story though, so I'll rename it then.

src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/builders/MatchingAttribute.java
```diff
@@ -0,0 +1,29 @@
+package uk.gov.ida.verifyserviceprovider.builders;
```
alan-gds: Seems weird to me for this to be in a 'builders' package.  I'll probably move it to a 'domain' package when I integrate this into the JSON schema story.

PR #113, opened 2018-11-28
FCLvYDk4: Handle non-matching non-success responses

alan-gds:
https://trello.com/c/FCLvYDk4/243-non-happy-path-for-non-matching-journey-in-vsp

Solo: @alan-gds

### DIFF
--- a/src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/NonMatchingAuthnFailedResponseAcceptanceTest.java
+++ b/src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/NonMatchingAuthnFailedResponseAcceptanceTest.java
@@ -51,7 +51,6 @@ public class NonMatchingAuthnFailedResponseAcceptanceTest {
         complianceTool.initialiseWithDefaultsForV2();
     }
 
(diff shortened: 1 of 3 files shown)


alan-gds:
We could do that.  I think they were done this way because that's how the matching ones were done.  I don't really have any strong feelings about it one way or the other, but if we refactor the non-matching ones we should probably refactor the matching ones too.

I think I'll let the second reviewer make the call, as I'm on the fence.

alan-gds:
No, the MockMsaServer is still doing the same job as before, mocking the MSA, so the name is still accurate.  The reason it's involved in the non-matching tests is because the VSP still depends on the MSA unconditionally for both versions of the journey, even though it doesn't really need it in the non-matching one.  Fixing that is a task in its own right, which we are planning to do later.

alan-gds:
I've given the suggestion of factoring out the setup boilerplate, but unfortunately I can't get it to work.  The client gets closed after one set of tests has run, and then any other tests that are using the shared code won't work; I can't see a way of fixing this.  So, I think we're stuck with the boilerplate unless anyone has any ideas.

PR #117, opened 2018-12-04
zjkU3Kyq: VSP to handle 'verified' field for gender

AmritSidhu:
https://trello.com/c/zjkU3Kyq/245-vsp-to-handle-verified-field-for-gender


jkunle:
I am in agreement with Alan, however you have captured the essence of this; so it shouldn't be too difficult to deal with. The conflict you'll need to deal with has to do with the Optionals. In some cases this has changed to non optional types on the DTO Andy's suggestions. Mainly because it affects the payload deserialisation. i.e. Other non-java platforms may be used with the VSP and would be expecting null values when there are empty values

PR #121, opened 2018-12-06
Y1qPhJZk: vsp should hash pids in non-matching-assertions

jkunle:
The VSP has the capability of Hashing PID for non matching journey,
however this functionality while it is present, is not currently used.
This PR address this by ensuring the non matching assertion service
performs hashing

-implemented test to show nonMatchingAssertionService performs hashing
-hashing functionality called from non assertion service

### DIFF
--- a/src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/NonMatchingAcceptanceTest.java
+++ b/src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/NonMatchingAcceptanceTest.java
@@ -56,7 +56,6 @@ public class NonMatchingAcceptanceTest {
 
     @Test
     public void shouldRespondWithIdentityVerifiedWhenVerificationSucceeds() {
-        String expectedPid = "some-expected-pid";
 
         complianceTool.initialiseWithDefaultsForV2();
 
--- a/src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/NonMatchingAcceptanceTest.java
+++ b/src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/NonMatchingAcceptanceTest.java
@@ -58,7 +58,7 @@ public class NonMatchingAcceptanceTest {
     public void shouldRespondWithIdentityVerifiedWhenVerificationSucceeds() {
         String expectedPid = "some-expected-pid";
 
-        complianceTool.initialiseWithPidForV2(expectedPid);
+        complianceTool.initialiseWithDefaultsForV2();
 
         RequestResponseBody requestResponseBody = generateRequestService.generateAuthnRequest(application.getLocalPort());
         Map<String, String> translateResponseRequestData = ImmutableMap.of(
@@ -77,7 +77,6 @@ public class NonMatchingAcceptanceTest {
 
         JSONObject jsonResponse = new JSONObject(response.readEntity(String.class));
         assertThat(jsonResponse.getString("scenario")).isEqualTo(IDENTITY_VERIFIED.name());
-        assertThat(jsonResponse.getString("pid")).isEqualTo(expectedPid);
         assertThat(jsonResponse.getString("levelOfAssurance")).isEqualTo(LEVEL_1.name());
     }
 
@@ -178,7 +177,6 @@ public class NonMatchingAcceptanceTest {
 
         JSONObject jsonResponse = new JSONObject(response.readEntity(String.class));
         assertThat(jsonResponse.getString("scenario")).isEqualTo(IDENTITY_VERIFIED.name());
-        assertThat(jsonResponse.getString("pid")).isEqualTo(expectedPid);
         assertThat(jsonResponse.getString("levelOfAssurance")).isEqualTo(LEVEL_1.name());
 
         JSONObject attributes = jsonResponse.getJSONObject("attributes");
--- a/src/main/java/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionService.java
+++ b/src/main/java/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionService.java
@@ -16,18 +16,18 @@ import uk.gov.ida.verifyserviceprovider.dto.NonMatchingAttributes;
 import uk.gov.ida.verifyserviceprovider.dto.NonMatchingScenario;
 import uk.gov.ida.verifyserviceprovider.dto.TranslatedNonMatchingResponseBody;
 import uk.gov.ida.verifyserviceprovider.exceptions.SamlResponseValidationException;
+import uk.gov.ida.verifyserviceprovider.factories.saml.UserIdHashFactory;
 import uk.gov.ida.verifyserviceprovider.mappers.MatchingDatasetToNonMatchingAttributesMapper;
 import uk.gov.ida.verifyserviceprovider.services.AssertionClassifier.AssertionType;
 import uk.gov.ida.verifyserviceprovider.validators.LevelOfAssuranceValidator;
 import uk.gov.ida.verifyserviceprovider.validators.SubjectValidator;
-
 import javax.xml.namespace.QName;
+import java.util.stream.Collectors;
 import java.util.Collection;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.stream.Collectors;
-
 import static java.util.Collections.singletonList;
 import static java.util.Optional.ofNullable;
 import static uk.gov.ida.saml.core.validation.errors.GenericHubProfileValidationSpecification.MISMATCHED_ISSUERS;
@@ -43,6 +43,7 @@ public class NonMatchingAssertionService implements AssertionService<TranslatedN
     private final AssertionClassifier assertionClassifierService;
     private final MatchingDatasetToNonMatchingAttributesMapper mdsMapper;
     private final LevelOfAssuranceValidator levelOfAssuranceValidator;
+    private UserIdHashFactory userIdHashFactory;
 
     public NonMatchingAssertionService(
             SamlAssertionsSignatureValidator assertionsSignatureValidator,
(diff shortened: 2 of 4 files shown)


src/main/java/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionService.java
```diff
@@ -72,7 +82,21 @@ public TranslatedNonMatchingResponseBody translateSuccessResponse(List<Assertion
 
         validate(authnAssertion, mdsAssertion, expectedInResponseTo, expectedLevelOfAssurance, levelOfAssurance);
 
+        validate(authnAssertion, mdsAssertion, expectedInResponseTo, expectedLevelOfAssurance, levelOfAssurance);
```
alan-gds: I'm not sure how this has happened but we now seem to be calling the validate method twice in a row! :)

src/main/java/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionService.java
```diff
@@ -72,7 +82,21 @@ public TranslatedNonMatchingResponseBody translateSuccessResponse(List<Assertion
...
         String nameID = mdsAssertion.getSubject().getNameID().getValue();
+        String issuerId =  authnAssertion.getIssuer().getValue();
+
+        String hashId = userIdHashFactory.hashId(
+                issuerId,
+                nameID,
+                ofNullable(authnStatement.getAuthnContext())
+                        .map(AuthnContext::getAuthnContextClassRef)
+                        .map(AuthnContextClassRef::getAuthnContextClassRef)
+                        .map(authnContextUriValue->getAuthnContextValueFromUri(authnContextUriValue))
```
alan-gds: This could be a method reference rather than a lambda, like on the previous lines.  It's probably worth changing that for consistency.

(IntelliJ pointed this out to me!)

src/main/java/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionService.java
```diff
@@ -72,7 +82,21 @@ public TranslatedNonMatchingResponseBody translateSuccessResponse(List<Assertion
...
         validate(authnAssertion, mdsAssertion, expectedInResponseTo, expectedLevelOfAssurance, levelOfAssurance);
 
+        validate(authnAssertion, mdsAssertion, expectedInResponseTo, expectedLevelOfAssurance, levelOfAssurance);
+
+        AuthnStatement authnStatement = authnAssertion.getAuthnStatements().get(0);
         String nameID = mdsAssertion.getSubject().getNameID().getValue();
+        String issuerId =  authnAssertion.getIssuer().getValue();
+
+        String hashId = userIdHashFactory.hashId(
+                issuerId,
```
alan-gds: I'd like to check with Andy, but I *think* this is wrong.  The issuerId here is the IDP's ID, whereas I think we want the RP's issuer ID.

src/main/java/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionService.java
```diff
@@ -192,4 +216,25 @@ private String extractLevelOfAssuranceStringFrom(Assertion authnAssertion) {
                 .map(AuthnContextClassRef::getAuthnContextClassRef)
                 .orElseThrow(() -> new SamlResponseValidationException("Expected a level of assurance."));
     }
+
+    private uk.gov.ida.saml.core.domain.AuthnContext getAuthnContextValueFromUri(String authnContextUriValue) {
```
alan-gds: I think it'd be worth moving this to a helper class of its own.  Also, it should be made static.

src/test/java/unit/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionServiceTest.java
```diff
@@ -233,6 +240,21 @@ public void shouldThrowExceptionWithUnknownLevelOfAssurance() throws Exception {
...
+
+        Assertion authnAssertion = anAuthnStatementAssertion(IdaAuthnContext.LEVEL_2_AUTHN_CTX, "requestId").buildUnencrypted();
+        Assertion mdsAssertion = aMatchingDatasetAssertionWithSignature(emptyList(), anIdpSignature(), "requestId").buildUnencrypted();
+
+        final String issuerId = authnAssertion.getIssuer().getValue();
+        final String nameId = authnAssertion.getSubject().getNameID().getValue();
+
+        nonMatchingAssertionService.translateSuccessResponse(ImmutableList.of(authnAssertion, mdsAssertion), "requestId", LEVEL_2, "default-entity-id");
+
+        verify(userIdHashFactory, times(1)).hashId(issuerId,nameId, Optional.of(AuthnContext.LEVEL_2));
```
alan-gds: As I've alluded to in another comment, rather than checking that the userIdHashFactory gets used, we should specify a return value for it and check that the response contains that value for the hash.

src/main/java/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionService.java
```diff
@@ -72,7 +82,21 @@ public TranslatedNonMatchingResponseBody translateSuccessResponse(List<Assertion
 
         validate(authnAssertion, mdsAssertion, expectedInResponseTo, expectedLevelOfAssurance, levelOfAssurance);
 
+        validate(authnAssertion, mdsAssertion, expectedInResponseTo, expectedLevelOfAssurance, levelOfAssurance);
+
+        AuthnStatement authnStatement = authnAssertion.getAuthnStatements().get(0);
         String nameID = mdsAssertion.getSubject().getNameID().getValue();
+        String issuerId =  authnAssertion.getIssuer().getValue();
+
+        String hashId = userIdHashFactory.hashId(
```
alan-gds: At least we are calculating the hashId now, but we are still not using it!  The problem is further down in the method, but Github won't let me put a comment there.  We are still using the raw nameId when we build the response body; instead, we should be using the hashId you have generated here.

We are also going to need to add a test which checks that we actually use the hashId; at the moment we only have one which checks that we generate it, which is why this hasn't been spotted.

src/main/java/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionService.java
jkunle: No longer required, I have changed the implementation

src/main/java/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionService.java
jkunle: You were right nameId now replaced by the calculated hashId, this also lead to failing acceptance tests; the ones that assert on pid so I removed these assert statements. As they are now covered by a new explicit acceptance test `responseShouldGenerateHashValueForSpecifiedPid`

src/main/java/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionService.java
jkunle: You may be right;  I have just implement this just as was done in MSA. It seems the idea of the above is that a different hashid depending on the selected IDP

alan-gds: Actually, you're right.  We do need to use the IDP's issuer ID, and the RP's issuer ID is already included, by virtue of it being passed into the UserIdHashFactory when it is created.  So, what you've got here is fine; sorry for muddying the waters!

src/main/java/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionService.java
```diff
@@ -73,10 +75,22 @@ public TranslatedNonMatchingResponseBody translateSuccessResponse(List<Assertion
...
 
         String nameID = mdsAssertion.getSubject().getNameID().getValue();
+        String issuerID = mdsAssertion.getIssuer().getValue();
+
+        String authnContextClassRef =  extractLevelOfAssuranceStringFrom(authnAssertion);
+
+        String hashId = userIdHashFactory.hashId(
+                issuerID,
+                nameID,
+                Arrays.stream(uk.gov.ida.saml.core.domain.AuthnContext.values())
```
alan-gds: Sorry to quibble over style but I would recommend using an import here rather than fully qualifying the class name, and I also think this section of code is complex enough that it's would be worth pulling it out into a local variable, rather than doing it all inside the method call brackets.  Obviously this isn't crucial to change though.

src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/NonMatchingAcceptanceTest.java
```diff
@@ -53,12 +55,32 @@ public static void setUpBeforeClass() {
...
+        Response response = client
+                .target(String.format("http://localhost:%d/translate-non-matching-response", application.getLocalPort()))
+                .request()
+                .buildPost(json(translateResponseRequestData))
+                .invoke();
+
+       JSONObject jsonResponse = new JSONObject(response.readEntity(String.class));
+
+        assertThat(response.getStatus()).isEqualTo(OK.getStatusCode());
+        assertThat(jsonResponse.getString("pid")).isEqualTo("56a33f20e83d6fb9a71cf962bd838d038dbdae5a5ee40dc098d6c62f7e2036d4");
```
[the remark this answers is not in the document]
jkunle: Yeah, I just ran that test and it failed. And just picked out the value it was expecting.

src/test/java/unit/uk/gov/ida/verifyserviceprovider/services/NonMatchingAssertionServiceTest.java
```diff
@@ -233,6 +243,27 @@ public void shouldThrowExceptionWithUnknownLevelOfAssurance() throws Exception {
         nonMatchingAssertionService.translateSuccessResponse(ImmutableList.of(authnAssertion, mdsAssertion), "requestId", LEVEL_2, "default-entity-id");
     }
 
+    @Test
+    public void expectedHashContainedInResponseBodyWhenUserIdFactoryIsCalledOnce() {
+
+        String requestId = "requestId";
+        String expected_hashed = "a5fbea969c3837a712cbe9e188804796828f369106478e623a436fa07e8fd298";
```
andy-paine: `expectedHashed`

src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/NonMatchingAcceptanceTest.java
```diff
@@ -56,9 +56,7 @@ public static void setUpBeforeClass() {
 
     @Test
     public void shouldRespondWithIdentityVerifiedWhenVerificationSucceeds() {
-        String expectedPid = "some-expected-pid";
 
-        complianceTool.initialiseWithPidForV2(expectedPid);
```
andy-paine: `complianceTool.initialiseWithDefaultsForV2();`

PR #122, opened 2018-12-14
NHstMcl3: Make MSA metadata optional

jkunle:
Making MSA data optional in the VSP configuration by making it nullable
in the case of non matching https://trello.com/c/NHstMcl3/242-make-msa-metadata-optional
Acceptance criteria is when all non-matching test that contain MSA configuration can
be made to pass without requiring MSA configuration data. Steps involved requires

- Removing multiple classes for non matching acceptance test
- This ensures there is only one place requiring MSA configuration data which can then be removed
- Making MSA metadata optional in VerifyServiceProvider configuration


src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/rules/VerifyServiceProviderAppRule.java
```diff
@@ -6,12 +6,27 @@
...
+            "verify-service-provider.yml",
+            ConfigOverride.config("serviceEntityIds", serviceEntityIdOverride),
+            ConfigOverride.config("hashingEntityId", "some-hashing-entity-id"),
+            ConfigOverride.config("server.connector.port", String.valueOf(0)),
+            ConfigOverride.config("logging.loggers.uk\\.gov", "DEBUG"),
+            ConfigOverride.config("samlSigningKey", TEST_RP_PRIVATE_SIGNING_KEY),
+            ConfigOverride.config("verifyHubConfiguration.environment", "COMPLIANCE_TOOL"),
+            ConfigOverride.config("samlPrimaryEncryptionKey", TEST_RP_PRIVATE_ENCRYPTION_KEY),
+            ConfigOverride.config("samlSecondaryEncryptionKey", secondaryEncryptionKey),
+            ConfigOverride.config("msaMetadata.uri", "http://some-msa-uri")
```
jkunle: `ConfigOverride.config("msaMetadata.uri", "http://some-msa-uri")` is this supposed to go too. I tried to remove this line and I started to get a null Exception

jkunle: So after discussing with Andy, this is currently out of scope for this card.

src/main/java/uk/gov/ida/verifyserviceprovider/configuration/VerifyServiceProviderConfiguration.java
```diff
@@ -23,7 +24,7 @@
     private PrivateKey samlSigningKey;
     private PrivateKey samlPrimaryEncryptionKey;
     private PrivateKey samlSecondaryEncryptionKey;
-    private MsaMetadataConfiguration msaMetadata;
+    private Optional<MsaMetadataConfiguration> msaMetadata;
```
alan-gds: I think I recall hearing Andy say we shouldn't use Optionals in Jackson-serialised classes, as it leads to having multiple ways to represent the absence of a value, which can get messy in Jackson-land.  Nulls are horrible but we're stuck with them in this context (I believe there are workarounds, but I don't know whether we're keen on using them).

Andy, please verify I'm not talking rubbish here - I wish we *could* use Optional!

src/main/java/uk/gov/ida/verifyserviceprovider/configuration/VerifyServiceProviderConfiguration.java
```diff
@@ -78,7 +79,7 @@ public PrivateKey getSamlSecondaryEncryptionKey() {
     }
 
     public MsaMetadataConfiguration getMsaMetadata() {
-        return msaMetadata;
+        return msaMetadata.orElseGet(()-> new MsaMetadataConfiguration(URI.create(""),0l, 0l, "",null,"",""));
```
alan-gds: Could we not just return null, rather than creating a dummy MsaMetadataConfiguration?

Or, we could even throw a RuntimeException!

src/main/java/uk/gov/ida/verifyserviceprovider/factories/VerifyServiceProviderFactory.java
```diff
@@ -138,10 +139,10 @@ private MetadataCredentialResolver getHubMetadataCredentialResolver() {
...
 
     private MetadataResolver getMsaMetadataResolver() {
-        return msaMetadataBundle.getMetadataResolver();
+        return msaMetadataBundle.map(item -> item.getMetadataResolver()).get();
     }
 
     private ExplicitKeySignatureTrustEngine getMsaSignatureTrustEngine() {
-        return msaMetadataBundle.getSignatureTrustEngine();
+        return msaMetadataBundle.map(item->item.getSignatureTrustEngine()).get();
     }
```
alan-gds: You could use method references instead of the two lambdas above, although it's all of a muchness really.  IntelliJ would prefer us to use method references though.

src/main/java/uk/gov/ida/verifyserviceprovider/configuration/VerifyServiceProviderConfiguration.java
jkunle: Thanks for your comment as always, I think in this case it may be fine because Jackson is used to deserialisation from verify-service-provider.yml to it's java equivalent VerifyServiceProviderConfiguration which in turn seems to be exclusively used by our code. So there are not external consumers who may assume an absence of data to be equivalent to NULL. So resource endpoints which have to return some data in JSON to be consumed by some client which may not necessarily be Java based would fall into this category. 

But I agree that there may be a case for being consistent. Depending on feedback from Andy, I would be happy to explore your workaround.

src/main/java/uk/gov/ida/verifyserviceprovider/configuration/VerifyServiceProviderConfiguration.java
jkunle: I think the idea is for this to work. i.e. just ignore the fact that this is missing; null would error out and a RuntimeException is also not desirable.

src/main/java/uk/gov/ida/verifyserviceprovider/factories/VerifyServiceProviderFactory.java
jkunle: Thanks for this. Will change this

PR #133, opened 2019-01-14
Write VSP release notes

bravegrape:
Write release notes for VSP release.
Also includes changes to config file comments to clarify the meaning of config settings.


verify-service-provider.yml
```diff
@@ -29,6 +29,8 @@ serviceEntityIds: ${SERVICE_ENTITY_IDS:-[]}
 # To use the Verify Service Provider with multiple services, add their entity IDs here, e.g.
 # '["service-one-entity-id", "service-two-entity-id"]'
 
+# This applies if you're connecting using only the VSP
+# TODO Explain what this is and how it should be configured
 hashingEntityId: ${HASHING_ENTITY_ID:-}
```
bravegrape: I need help explaining this bit 🙈

RELEASE_NOTES.md
```diff
@@ -3,8 +3,33 @@ Release notes
...
+
+We moved the [VSP tutorial](https://www.docs.verify.service.gov.uk/get-started-with-vsp/#get-started-with-the-verify-service-provider) to the new technical documentation website and updated it to help you connect to GOV.UK verify using the VSP.
+
+#### Support for compilation on Java 9 and 10
+
+Compilation on Java 9 or 10 will produce a multi-release JAR that will run on Java 8, 9, and 10. This release was compiled on Java 8 and does not contain a multi-release JAR.
+
+#### Metadata health check changes
+
+Changed **HOW?** metadata health check names to use the URI of the metadata they are trying to resolve
```
bravegrape: How were the metadata health check names changed and how does this affect users?

verify-service-provider.yml
```diff
@@ -29,6 +29,8 @@ serviceEntityIds: ${SERVICE_ENTITY_IDS:-[]}
 # To use the Verify Service Provider with multiple services, add their entity IDs here, e.g.
 # '["service-one-entity-id", "service-two-entity-id"]'
 
+# This applies if you're connecting using only the VSP
```
alan-gds: Also, it's optional if you only have one service (i.e. only one entry in serviceEntityIds above).  It's only multi-tenanted RPs who need to configure a hashingEntityId.

RELEASE_NOTES.md
```diff
@@ -3,8 +3,33 @@ Release notes
 
 ### Next
 
-* Included support for compilation on Java 9 and 10. Compilation on Java 9 or 10 will produce a multi-release JAR that will run on Java 8, 9, and 10. This release was compiled on Java 8 and does not contain a multi-release JAR.
-* Changed metadata health check names to use the URI of the metadata they are trying to resolve
+#### Connect to Verify using only the VSP
+
+This release adds the ability to connect to GOV.UK Verify using only the Verify Service Provider (VSP). This means services can connect without needing to host a [Matching Service Adapter (MSA)](https://github.com/alphagov/verify-matching-service-adapter). This makes it easier to connect to Verify and reduces maintenance tasks once connected.
+
+When used alone, the VSP provides MSA functionality. It can:
```
alan-gds: That might be a bit misleading.  The core job of the MSA was to talk to the RP's Local Matching Service, which the VSP will not be doing.

RELEASE_NOTES.md
alan-gds: Someone needs to dig into the history and either find out what was done, or find out how did it and ask them.  I'd prefer not to be sucked into this while I'm on the yak desk, so please ask another dev on the team.

RELEASE_NOTES.md
```diff
@@ -3,8 +3,33 @@ Release notes
...
+
+For more information on connecting to GOV.UK Verify using only the VSP, see the [technical documentation](https://www.docs.verify.service.gov.uk).
+
+#### Tutorial for the new VSP
+
+We moved the [VSP tutorial](https://www.docs.verify.service.gov.uk/get-started-with-vsp/#get-started-with-the-verify-service-provider) to the new technical documentation website and updated it to help you connect to GOV.UK verify using the VSP.
+
+#### Support for compilation on Java 9 and 10
+
+Compilation on Java 9 or 10 will produce a multi-release JAR that will run on Java 8, 9, and 10. This release was compiled on Java 8 and does not contain a multi-release JAR.
```
alan-gds: Was this copied from somewhere?  I don't think it's up to date.  Make sure it matches up with the expectation of whoever is doing the release - we will either build against Java 8 as stated, or build a multi-release JAR against Java 11.  I'm afraid I'm not likely to have time to dig into this myself.  It would be worth talking to Irina about this.

verify-service-provider.yml
```diff
@@ -29,6 +29,8 @@ serviceEntityIds: ${SERVICE_ENTITY_IDS:-[]}
...
+# TODO Explain what this is and how it should be configured
```
alan-gds: Best to ask a tech arch about this (John or Chris).  From my perspective, it's just an arbitrary string; I don't know what we recommend.

verify-service-provider.yml
```diff
@@ -55,6 +58,9 @@ samlPrimaryEncryptionKey: ${SAML_PRIMARY_ENCRYPTION_KEY:-}
 # This only needs to be set if during key rotations (for example if your primary encryption certificate is about to expire)
 samlSecondaryEncryptionKey: ${SAML_SECONDARY_ENCRYPTION_KEY:-}
 
+# Enable European identities
+# Setting `europeanIdentity.enabled: true` allows your VSP to consume European
+# identities.
```
alan-gds: Please note that the whole section is now optional.  If it is omitted, it will not be possible to process eIDAS identities.  Likewise, if the section is present but the 'enabled' setting is false, the other settings will not be used (although if they are present, their values must be valid) and it will not be possible to process eIDAS identities.

RELEASE_NOTES.md
bravegrape: It looks like the content I put together for an MSA release. I'll make sure this matches reality 👌

verify-service-provider.yml
bravegrape: I see. I've re-organised comments to make this clearer :)

RELEASE_NOTES.md
alan-gds: Andy is your best bet for this, I think (or perhaps Chris H).

verify-service-provider.yml
alan-gds: Chris H would be a good person to ask.

RELEASE_NOTES.md
```diff
@@ -3,8 +3,38 @@ Release notes
...
+
+If you are already connected to GOV.UK Verify and are running an MSA, you can upgrade to VSP 2.0.0 without needing to do any configuration changes.
+
+If you want to switch off your Matching Service Adapter and only use the Verify Service Provider to connect to Verify, [contact the GOV.UK Verify Team](https://www.verify.service.gov.uk/support/).
+
+For more information on connecting to GOV.UK Verify using only the VSP, see the [technical documentation](https://www.docs.verify.service.gov.uk).
+
+#### Improved technical documentation
+
+The [new technical documentation website](https://www.docs.verify.service.gov.uk) has new guidance to help users connect to GOV.UK Verify using the VSP.
```
kerrr: The _new_ technical ... has _new_ guidance ...
Would this read better with one of the 'new's omitted or changed up? Maybe switch the first one for 'improved'?

verify-service-provider.yml
```diff
@@ -55,7 +58,12 @@ samlPrimaryEncryptionKey: ${SAML_PRIMARY_ENCRYPTION_KEY:-}
...
 samlSecondaryEncryptionKey: ${SAML_SECONDARY_ENCRYPTION_KEY:-}
 
+# Enable European identities (optional)
+# Include this section to enable your VSP to process European identities.
 europeanIdentity:
+  # `enabled: true` to allow your VSP to process European identities
+  # `enabled: false` means your VSP won't process European identities, but
+  # the section must contain valid key-value pairs.
   enabled: ${EUROPEAN_IDENTITY_ENABLED}
   hubConnectorEntityId: ${HUB_CONNECTOR_ENTITY_ID}
```
kerrr: Most of this can be configured by setting the Hub Environment now.

kerrr: Most of this can be configured with the Hub Environment now.

verify-service-provider.yml
jakubmiarka: This is for generating the PID. It uses a hashing algorithm in combination with the `hashingEntityId` to make sure the same user has the same PID (i.e. is always hashed the same way) and services can match them. As Alan mentioned above, it's only required if it's a multi-tenant set-up, otherwise it will use the serviceEntityId by default.

verify-service-provider.yml
jakubmiarka: see my comment above

verify-service-provider.yml
jakubmiarka: That's right. The section now only requires the `enabled` flag. The rest is now automagically configured (but it can still be overwritten)

RELEASE_NOTES.md
```diff
@@ -3,8 +3,38 @@ Release notes
...
+* implement security features to the same level as the MSA
+* handle both GOV.UK Verify and European identities
+* do Personal Identifier (PID) hashing
+
+If you are already connected to GOV.UK Verify and are running an MSA, you can upgrade to VSP 2.0.0 without needing to do any configuration changes.
+
+If you want to switch off your Matching Service Adapter and only use the Verify Service Provider to connect to Verify, [contact the GOV.UK Verify Team](https://www.verify.service.gov.uk/support/).
+
+For more information on connecting to GOV.UK Verify using only the VSP, see the [technical documentation](https://www.docs.verify.service.gov.uk).
+
```
chrisholmes: I think we need at a caveat here. Something like:
```suggestion
##### Caveats

It is not currently possible to test europeanIdentity with the VSP and the Compliance Tool. We recommend that you disable europeanIdentities support when you set the HUB_ENVIRONMENT configuration option to 'COMPLIANCE_TOOL'
```

RELEASE_NOTES.md
bravegrape: I agree. Changed to say 'the required MSA functionality'.

RELEASE_NOTES.md
bravegrape: Spoke to Andy and it's all sorted now 👍

RELEASE_NOTES.md
bravegrape: I've now removed this completely, following our conversation.
Here's the commit: https://github.com/alphagov/verify-service-provider/pull/133/commits/3063e3921a742d7bd47642c02e660dd52014c294

RELEASE_NOTES.md
bravegrape: I've rephrased :)

RELEASE_NOTES.md
bravegrape: Following our chat, I won't be adding this comment in, because new users connecting with VSP 2.0.0 and above will not be using `HUB_ENVORONMENT` set to `COMPLIANCE_TOOL`. They'll be using the development command which automatically does the configuration and does not cause errors.
Follow up: remove `COMPLIANCE_TOOL` from the list of recommended values for `HUB_ENVIRONMENT` in the config.

verify-service-provider.yml
bravegrape: Cool! Leaving only the `enabled` flag 👍

verify-service-provider.yml
```diff
@@ -24,23 +22,33 @@ logging:
...
 # '["service-one-entity-id", "service-two-entity-id"]'
+serviceEntityIds: ${SERVICE_ENTITY_IDS:-[]}
 
+# Use `hashingEntityId` if you're connecting
+# more than one service to a single VSP deployment.
+#
+# The VSP uses the `hashingEntityId` instead of the value in `serviceEntityIds`
+# when generating a user's hashed personal identifier (PID). This is to make
+# sure a user verified by a particular identity provider is linked to a unique
+# hashed PID no matter which one of your services they're requesting to use.
```
jakubmiarka: nice! 👍

PR #134, opened 2019-01-16
Pi1Z0jDx - eIDAS flagging implementation

jkunle:


src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/NonMatchingAcceptanceTest.java
```diff
@@ -38,7 +40,24 @@
 public class NonMatchingAcceptanceTest {
 
     @ClassRule
-    public static VerifyServiceProviderAppRule application = new VerifyServiceProviderAppRule();
+    public static MockMsaServer msaServer = new MockMsaServer();
+
+    @ClassRule
+    public static VerifyServiceProviderAppRule application = aVerifyServiceProviderAppRule()
```
chrisholmes: 👍 on making this a builder

src/main/java/uk/gov/ida/verifyserviceprovider/services/EidasAssertionService.java
```diff
@@ -88,7 +92,7 @@ public LevelOfAssurance extractLevelOfAssuranceFrom(Assertion countryAssertion)
     }
 
     public Boolean isCountryAssertion(Assertion assertion) {
-        return metadataResolverRepository.getResolverEntityIds().contains(assertion.getIssuer().getValue());
+        return isEnabled && metadataResolverRepository.get().getResolverEntityIds().contains(assertion.getIssuer().getValue());
```
chrisholmes: I don't think this should block the PR, but I think we could use something similar to the [nullobject pattern](https://en.wikipedia.org/wiki/Null_object_pattern) here so that we don't need to pass Optional arguments and booleans everywhere.

src/main/java/uk/gov/ida/verifyserviceprovider/factories/VerifyServiceProviderFactory.java
```diff
@@ -144,20 +147,29 @@ private ExplicitKeySignatureTrustEngine getMsaSignatureTrustEngine() {
...
-            configuration.getEuropeanIdentity().getAggregatedMetadata(),
-            new DropwizardMetadataResolverFactory(),
-            new Timer(),
-            new MetadataSignatureTrustEngineFactory(),
-            new MetadataResolverConfigBuilder(),
-            client
-        );
+    private Optional<EidasMetadataResolverRepository> getEidasMetadataResolverRepository() {
+        if (isEidasEnabled()) {
+            return Optional.of(new EidasMetadataResolverRepository(
```
chrisholmes: please see later comment about use of the null object pattern

PR #136, opened 2019-01-17
7DyD3EvQ: implement a compliance tool mode for the VSP

chrisholmes:
WHY
====

Interacting with the Compliance Tool (CT) is not a straightforward process and the API could be better.
When a user is using the VSP for the first time it can be frustrating having to make sense of the CT API in order to make progress.
In order to get the VSP running with the CT so that can start development they need to:
1. Create a set of private keys and certificates
2. Configure the VSP to interact with the CT and use the newly generated keys
   and certificates
3. Create a script that can initiate a session with the CT
4. Create a client that can interact with the VSP
5. Test the interaction between the client, VSP, and CT

# WHAT

The VSP will provide a development mode initialized by a Dropwizard command.

When starting in this mode the VSP will:
- create in-memory keys and self-sign certificates
- use a random UUID for an entity ID
- configure itself to read the CT's metadata
- configure itself to run in non-matching mode
- initialize a new session with the compliance tool. 

A user of the VSP will be able start development mode by running the following
command from the shell:

```
./bin/verify-service-provider development
```

The user should also be able to run this mode on Windows.

The command will provide additional options to the user so that they can
control:
- the host the VSP will bind to
- the port the VSP will run on
- the location where the CT will send SAML responses to
- the contents of the matching dataset that the CT will use in SAML responses

In order to simplify the implementation of this mode the behaviour will only be
available to users of the 'non-matching' journey.

# Consequences

- Supersedes [ADR-0008](docs/adr/0008-provide-an-end-to-end-stub.md) (which was
    never implemented)
- The VSP will depend on bouncycastle to create the in-memory certificates

# Caveats

When the VSP runs it will always immediately write the following to stderr:

```
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/Users/christopherholmes/.gradle/caches/modules-2/files-2.1/uk.gov.ida.eidas/trust-anchor/1.0-51/4ed2e2e7c79f0bec6eec362de078a065cf8d7f20/trust-anchor-1.0-51.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/Users/christopherholmes/.gradle/caches/modules-2/files-2.1/ch.qos.logback/logback-classic/1.2.3/7c4f3c474fb2c041d8028740440937705ebb473a/logback-classic-1.2.3.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [ch.qos.logback.classic.util.ContextSelectorStaticBinder]
```
This makes it harder to use the VSP as a command line options as any messages (like usage) have to follow the error message.


jkunle:
This is very good stuff. The classes in compliance Domain seems to be duplicates; maybe they be inherited that way if there are changes to the original it would break the compliance ones. As it stands there is a chance of missing these changes I am happy to approve this if you consider this a minor issue

chrisholmes:
> This is very good stuff. The classes in compliance Domain seems to be duplicates; maybe they be inherited that way if there are changes to the original it would break the compliance ones. As it stands there is a chance of missing these changes I am happy to approve this if you consider this a minor issue

I agree there's a bit of duplication now in what should be dto objects. If you're fine with it I will hold off on the deduplication until after we merge/close this PR as the changes are already pretty big.

bravegrape:
'The user should also be able to run this mode on Windows.'

@chrisholmes Just to clarify for tech docs purposes, will Windows users be able to run this or does it still need testing?

chrisholmes:
> 'The user should also be able to run this mode on Windows.'
> 
> @chrisholmes Just to clarify for tech docs purposes, will Windows users be able to run this or does it still need testing?

@bravegrape I haven't been able to test this yet but it should work so long as the VSP can still run on Windows

PR #139, opened 2019-01-23
BgwZOwlj: Add tests for hashingEntityId

alan-gds:
https://trello.com/c/BgwZOwlj/41-make-hashingentityid-optional

The logic around whether hashingEntityId is allowed to be absent in the config is just about complex enough that we should have some tests for it.  Some of the existing tests kind of happen to cover that things work correctly when it is present; I have added some tests which check that we get the correct behaviour when it's not present.

I'm not great with DropWizard AppRules, and I can't see a clean way to get the tests into the same class file, so I've ended up with a bit of a weird test class which only tests the absence cases, rather than having all of the relevant test coverage in one place - suggestions for how to clean this up are welcome.  Likewise, the only way I know of to get rid of a config item in an AppRule is to base the AppRule on a config file which doesn't have the config item - setting it to null in a ConfigOverride doesn't seem to work.  Again, if you know a cleaner way to do this, I am all ears!


alan-gds:
We already had better test coverage for this elsewhere; I am a muppet.

PR #158, opened 2019-02-14
EID-1325: Add RequestedAuthnContext to AuthnRequest

vixus0:
The `levelOfAssurance` parameter sent to the `/generate-request` endpoint
was being ignored. This change adds the corresponding AuthnContextClassRef(s) to
the RequestedAuthnContext of the AuthnRequest generated by the VSP.

In the case of LOA1, we need to add both the level 1 and level 2
AuthnContextClassRefs while for LOA2 we only need the level 2
AuthnContextClassRef.

### DIFF
--- a/src/main/java/uk/gov/ida/verifyserviceprovider/factories/saml/AuthnRequestFactory.java
+++ b/src/main/java/uk/gov/ida/verifyserviceprovider/factories/saml/AuthnRequestFactory.java
@@ -5,14 +5,19 @@ import org.opensaml.core.xml.config.XMLObjectProviderRegistrySupport;
 import org.opensaml.core.xml.io.MarshallingException;
 import org.opensaml.saml.common.SAMLRuntimeException;
 import org.opensaml.saml.saml2.core.Attribute;
+import org.opensaml.saml.saml2.core.AuthnContextClassRef;
+import org.opensaml.saml.saml2.core.AuthnContextComparisonTypeEnumeration;
 import org.opensaml.saml.saml2.core.AuthnRequest;
 import org.opensaml.saml.saml2.core.EncryptedAttribute;
 import org.opensaml.saml.saml2.core.Extensions;
 import org.opensaml.saml.saml2.core.Issuer;
+import org.opensaml.saml.saml2.core.RequestedAuthnContext;
 import org.opensaml.saml.saml2.core.impl.AttributeBuilder;
+import org.opensaml.saml.saml2.core.impl.AuthnContextClassRefBuilder;
 import org.opensaml.saml.saml2.core.impl.AuthnRequestBuilder;
 import org.opensaml.saml.saml2.core.impl.ExtensionsBuilder;
 import org.opensaml.saml.saml2.core.impl.IssuerBuilder;
+import org.opensaml.saml.saml2.core.impl.RequestedAuthnContextBuilder;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/ida/verifyserviceprovider/factories/saml/AuthnRequestFactory.java
```diff
@@ -84,6 +103,18 @@ public AuthnRequest build(LevelOfAssurance levelOfAssurance, String serviceEntit
...
 
+    private RequestedAuthnContext createRequestedAuthnContext(List<String> authnContexts) {
+        RequestedAuthnContext requestedAuthnContext= new RequestedAuthnContextBuilder().buildObject();
+        List<AuthnContextClassRef> authnContextClassRefs = authnContexts.stream().map((a) -> {
+            AuthnContextClassRef authnContextClassRef = new AuthnContextClassRefBuilder().buildObject();
+            authnContextClassRef.setAuthnContextClassRef(a);
+            return authnContextClassRef;
+        }).collect(Collectors.toList());
+        requestedAuthnContext.getAuthnContextClassRefs().addAll(authnContextClassRefs);
+        requestedAuthnContext.setComparison(AuthnContextComparisonTypeEnumeration.EXACT);
```
chrisholmes: Is this the right compaision type? I thought it was supposed be `minimum`. Is there a spec somewhere from when we decided this 2(?) years ago.

src/main/java/uk/gov/ida/verifyserviceprovider/factories/saml/AuthnRequestFactory.java
```diff
@@ -61,12 +70,22 @@ public AuthnRequestFactory(
     }
 
     public AuthnRequest build(LevelOfAssurance levelOfAssurance, String serviceEntityId) {
+        List<String> authnContexts = new ArrayList<>();
+        switch (levelOfAssurance) {
+            case LEVEL_1:
+                authnContexts.add(IdaAuthnContext.LEVEL_1_AUTHN_CTX);
```
chrisholmes: If we're using 'exact' then 'technically' I think we need to include LEVEL_2 aswell.

src/main/java/uk/gov/ida/verifyserviceprovider/factories/saml/AuthnRequestFactory.java
vixus0: `minimum` was the legacy comparison type and `exact` the new one. At the moment, only IDPs care about the comparison type, hence why we added a couple of new IDP tests to compliance tool to test it (see commit `fc58f88a01ab8f100122337b67520c68991782c8`).

src/main/java/uk/gov/ida/verifyserviceprovider/factories/saml/AuthnRequestFactory.java
vixus0: The `switch` fall through should take care of this. See the two new tests in `AuthnRequestFactoryTest`.

src/main/java/uk/gov/ida/verifyserviceprovider/factories/saml/AuthnRequestFactory.java
chrisholmes: The reason I ask is I remember us saying that we want to use a different comparison type for the SPs and IDPs.

PR #164, opened 2019-02-20
 zEyIoEnS: Return most attributes with their historical values

chrisholmes:
The SAML profile specified that attribtue should be provided as a list with historical context, but we were ignoring this and only returning the first of some attributes (firstName, and Date Of Birth). In some case this could cause problems if a non-verified attribute was returned when a verified attribute was available.

This PR changes the attributes returned to make them more consistent with the SAML profile.
The following attributes will now be provided with possible historical values:
- firstNames
- middleNames
- surnames
- dates Of Birth
- addresses

Gender will still be provided as a single attribute.

### DIFF
--- a/src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/Utils/MdsValueChecker.java
+++ b/src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/Utils/MdsValueChecker.java
@@ -5,6 +5,7 @@ import org.json.JSONArray;
 import org.json.JSONObject;
 
 import java.util.List;
+import java.util.Optional;
 
 public class MdsValueChecker {
 
@@ -33,8 +34,8 @@ public class MdsValueChecker {
     public static void checkMdsValueOfAddress(
             int index,
             List<String> lines,
-            String postcode,
-            String internationalPostcode,
+            Optional<String> postcode,
+            Optional<String> internationalPostcode,
             boolean expectedIsVerified,
             String expectedFromDateString,
             String expectedToDateString,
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/ida/verifyserviceprovider/dto/NonMatchingAttributes.java
```diff
@@ -70,20 +70,20 @@ public boolean equals(Object o) {
         if (o == null || getClass() != o.getClass()) return false;
 
         NonMatchingAttributes that = (NonMatchingAttributes) o;
-        if (firstName != null ? !firstName.equals(that.firstName) : that.firstName != null) return false;
+        if (firstNames != null ? !firstNames.equals(that.firstNames) : that.firstNames != null) return false;
```
kerrr: All this could be simplified with:
```java
        return Objects.equals(firstNames, that.firstNames) &&
                Objects.equals(middleNames, that.middleNames) &&
                Objects.equals(surnames, that.surnames) &&
                Objects.equals(datesOfBirth, that.datesOfBirth) &&
                Objects.equals(gender, that.gender) &&
                Objects.equals(addresses, that.addresses);
```

src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/Utils/MdsValueChecker.java
```diff
@@ -51,8 +52,17 @@ public static void checkMdsValueOfAddress(
         for (index = 0; index < lines.size(); index++) {
             Assertions.assertThat(jsonLines.getString(index)).isEqualTo(lines.get(index));
         }
-        Assertions.assertThat(addressValue.getString("postCode")).isEqualTo(postcode);
-        Assertions.assertThat(addressValue.getString("internationalPostCode")).isEqualTo(internationalPostcode);
+        if(postcode.isPresent()) {
+            Assertions.assertThat(addressValue.getString("postCode")).isEqualTo(postcode.get());
+        } else {
+
```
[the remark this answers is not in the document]
chrisholmes: I will address the clean-up of the address assertions in a follow-up PR when we add the UPRN into the response.

src/main/java/uk/gov/ida/verifyserviceprovider/dto/NonMatchingAttributes.java
chrisholmes: Yeah, that's true. This is auto-generated by Intellij so I'm going to leave it alone now.

PR #165, opened 2019-02-21
Allow external binaries in docker build

blairboy362:
In some circumstances the artifactory repo isn't available (proxy-node) and so we need the option to fall back to the public ones.

Solo: @blairboy362

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,7 @@ data/**
 /.mongo/*
 !/.mongo/.gitkeep
 node_modules
+.classpath
+.project
+.settings
+bin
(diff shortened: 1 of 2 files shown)


andy-paine:
ARG or ENV? 🤔

blairboy362:
@andy-paine ARG. We don't want it to persist beyond build stage...

https://docs.docker.com/engine/reference/#scope

andy-paine:
I was just confused as to how that `ARG` got provided to gradle so it knew to use the public ones?

blairboy362:
@andy-paine I'm hoping via...
https://github.com/alphagov/verify-eidas-pipelines/blob/master/templates/build.yaml#L450

PR #166, opened 2019-02-21
OLqSibaL: Add UPRN to Addresses returned by the VSP

chrisholmes:
# What
Addresses from our IDPs may include a UPRN field, but the VSP is currently ignoring it in our response back to the service. We should include it as it may be helpful to a service.

# Why
The UPRN could be used by a service to assist with matching or other reasons.

### DIFF
--- a/src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/Utils/MdsValueChecker.java
+++ b/src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/Utils/MdsValueChecker.java
@@ -3,7 +3,9 @@ package uk.gov.ida.verifyserviceprovider.Utils;
 import org.assertj.core.api.Assertions;
 import org.json.JSONArray;
 import org.json.JSONObject;
+import uk.gov.ida.verifyserviceprovider.compliance.dto.MatchingAddress;
 
+import java.time.format.DateTimeFormatter;
 import java.util.List;
 import java.util.Optional;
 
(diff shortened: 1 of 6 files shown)


src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/Utils/MdsValueChecker.java
```diff
@@ -31,15 +33,39 @@ public static void checkMdsValueOfAttributeWithoutDates(
         checkMdsValueInJsonObject(attribute, expectedValue, expectedIsVerified);
     }
 
+    public static void checkMdsValueOfAddress(
+            int index,
+            JSONObject attributes,
+            MatchingAddress matchingAddress
```
alan-gds: It'd be worth switching round the attributes and matchingAddress parameters, so that they're in line with the order used in other MdsValueChecker methods.  For the same reason, I'd recommend renaming matchingAddress to expectedAddress.

src/main/java/uk/gov/ida/verifyserviceprovider/dto/NonMatchingAddress.java
```diff
@@ -44,6 +50,7 @@ public boolean equals(Object o) {
 
         if (lines != null ? !lines.equals(other.lines) : other.lines != null) return false;
         if (postCode != null ? !postCode.equals(other.postCode) : other.postCode != null) return false;
+        if (uprn != null ? !uprn.equals(other.uprn) : other.uprn != null) return false;
```
alan-gds: The hashCode method needs to be updated! :O

PR #167, opened 2019-02-22
gUPcguib: Use LocalDate instead of LocalDateTime

alan-gds:
The from/to dates in NonMatchingVerifiableAttribute should be LocalDates, not LocalDateTimes.  The time component is not useful.

### DIFF
--- a/src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/NonMatchingAcceptanceTest.java
+++ b/src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/NonMatchingAcceptanceTest.java
@@ -155,11 +155,11 @@ public class NonMatchingAcceptanceTest {
         JSONObject attributes = jsonResponse.getJSONObject("attributes");
         assertThat(attributes.keys()).containsExactlyInAnyOrder(COMMON_ATTRIBUTES);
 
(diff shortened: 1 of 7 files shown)


src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/NonMatchingAcceptanceTest.java
```diff
@@ -224,7 +224,7 @@ public void shouldCorrectlyHandleEmptyValuesInMatchingDataset() {
 
         JSONObject attributes = jsonResponse.getJSONObject("attributes");
         assertThat(attributes.keys()).containsExactlyInAnyOrder(COMMON_ATTRIBUTES);
-        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd");
         String expectedFromDateString = standardFromDate.toLocalDate().atStartOfDay().format(formatter).replace(" ", "T");
```
chrisholmes: We don't need these line string replacement lines anymore.

alan-gds: Good point.  I've updated the PR (via an amend/force push, because I also noticed I'd forgotten to put the Trello link in the original commit message).

PR #169, opened 2019-02-22
9ZNOyJJW: include nonLatinScriptValue on name attributes

chrisholmes:
This PR adds support for serializing the `nonLatinScriptValue` from eIDAS Assertions into the attributes returned from the VSP.

### DIFF
--- a/src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/NonMatchingEidasAcceptanceTest.java
+++ b/src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/NonMatchingEidasAcceptanceTest.java
@@ -120,16 +120,16 @@ public class NonMatchingEidasAcceptanceTest {
     public void shouldMapAttributesCorrectly() throws Exception {
         AttributeStatementBuilder attributeStatementBuilder = AttributeStatementBuilder.anAttributeStatement();
         
-        Attribute firstName =  anAttribute(IdaConstants.Eidas_Attributes.FirstName.NAME);
+        Attribute givenName =  anAttribute(IdaConstants.Eidas_Attributes.FirstName.NAME);
         CurrentGivenName firstNameValue = new CurrentGivenNameBuilder().buildObject();
         firstNameValue.setFirstName("Joe");
-        firstName.getAttributeValues().add(firstNameValue);
+        givenName.getAttributeValues().add(firstNameValue);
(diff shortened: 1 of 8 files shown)


src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/NonMatchingEidasAcceptanceTest.java
```diff
@@ -28,7 +43,17 @@
...
 
-     @Test
+    private OpenSamlXmlObjectFactory openSamlXmlObjectFactory;
+
+    @Before
+    public void setUp() throws Exception {
+        openSamlXmlObjectFactory = new OpenSamlXmlObjectFactory();
+    }
+
+    @Before
```
alan-gds: This @Before is floating around on its own and looks like it shouldn't be there at all.

src/main/java/uk/gov/ida/verifyserviceprovider/mappers/MatchingDatasetToNonMatchingAttributesMapper.java
```diff
@@ -53,18 +54,30 @@ public NonMatchingAttributes mapToNonMatchingAttributes(MatchingDataset matching
...
                 .sorted(attributeComparator())
                 .collect(Collectors.toList());
     }
 
 
-    public static <T> Comparator<NonMatchingVerifiableAttribute<T>> attributeComparator() {
-        return Comparator.<NonMatchingVerifiableAttribute<T>, LocalDateTime>comparing(NonMatchingVerifiableAttribute::getTo, Comparator.nullsFirst(Comparator.reverseOrder()))
-                .thenComparing(NonMatchingVerifiableAttribute::isVerified, Comparator.reverseOrder())
-                .thenComparing(NonMatchingVerifiableAttribute::getFrom, Comparator.nullsLast(Comparator.reverseOrder()));
+    private NonMatchingTransliterableAttribute mapToNonMatchingVerifiableAttribute(TransliterableMdsValue transliterableMdsValue) {
```
alan-gds: This method name seems misleading.

chrisholmes:
> Basically looks okay aside from some small things I've put comments on. Also, would it be worth having a test which demonstrates what we do if we received non-latin script values when eIDAS is not enabled?

I think that might be a bit of a tricky test to write. Would this be if we receive an eIDAS-format attribute in a Verify-format Assertion? That would be an interesting test to add, but I believe it's bigger than the scope of this story

PR #178, opened 2019-03-01
iziGhhu2: Hash eIDAS PIDs

jakubmiarka:


src/test/java/unit/uk/gov/ida/verifyserviceprovider/services/EidasAssertionServiceTest.java
```diff
@@ -89,7 +92,8 @@ public void setUp() {
             levelOfAssuranceValidator,
             Optional.of(metadataResolverRepository),
             signatureValidatorFactory,
-            Optional.of(HUB_CONNECTOR_ENTITY_ID));
+            Optional.of(HUB_CONNECTOR_ENTITY_ID),
+            userIdHashFactory);
```
chrisholmes: Could we get a test for this? Also, something in an integration test for it?

src/main/java/uk/gov/ida/verifyserviceprovider/services/EidasAssertionService.java
```diff
@@ -32,6 +33,8 @@
     private final Optional<EidasMetadataResolverRepository> metadataResolverRepository;
     private final SignatureValidatorFactory signatureValidatorFactory;
     private final Optional<String> entityId;
+    private UserIdHashFactory userIdHashFactory;
+    private final AuthnContextFactory authnContextFactory = new AuthnContextFactory();
```
chrisholmes: We should probably DI this, but we can clean it up later.

src/main/java/uk/gov/ida/verifyserviceprovider/services/EidasAssertionService.java
```diff
@@ -95,6 +105,12 @@ public LevelOfAssurance extractLevelOfAssuranceFrom(Assertion countryAssertion)
         }
     }
 
+    private Optional<AuthnContext> getAuthnContext(String uri) {
+        AuthnContext authnContext = authnContextFactory.mapFromEidasToLoA(uri);
+
+        return Optional.of(authnContext);
```
chrisholmes: This will through an Exception if the LOA is null

jakubmiarka: Yep, it should not be null at this point...  moreover the hashing actually requires LOA (AuthnContext) and will throw an exception too anyway, so I'm not 100% sure why we're passing this around as optional 🤷‍♂️

src/test/java/unit/uk/gov/ida/verifyserviceprovider/services/EidasAssertionServiceTest.java
```diff
@@ -120,7 +113,7 @@ public void shouldCallValidatorsCorrectly() {
         List<Assertion> assertions = asList(
             anAssertionWithAuthnStatement(EIDAS_LOA_HIGH, "requestId").buildUnencrypted());
 
-        eidasAssertionService.translateSuccessResponse(assertions, "requestId", LevelOfAssurance.LEVEL_2, null);
+        eidasAssertionService.translateSuccessResponse(assertions, "requestId", LEVEL_2, null);
```
alan-gds: I disagree with Codacy that this is an improvement, but it's probably not a battle worth fighting.

PR #179, opened 2019-03-04
bau: make loa request param optional

chrisholmes:
This parameter isn't used for anything so we can make it optional with
a view to phasing it out (or doing something with it) in the next
release.

It should be able to handle empty request bodies, '{}', and
'{"entityId": "foobar"}' bodies.

Also, removed the passing of 'levelOfAssurance' to objects where it's
never used.

Note: we've removed this parameter from our docs so if we don't remove it now then we need to fix the docs.

### DIFF
--- a/src/test/java/unit/uk/gov/ida/verifyserviceprovider/resources/GenerateAuthnRequestResourceTest.java
+++ b/src/test/java/unit/uk/gov/ida/verifyserviceprovider/resources/GenerateAuthnRequestResourceTest.java
@@ -1,5 +1,6 @@
 package unit.uk.gov.ida.verifyserviceprovider.resources;
 
+import com.google.common.collect.ImmutableMap;
 import io.dropwizard.jersey.errors.ErrorMessage;
 import io.dropwizard.testing.junit.ResourceTestRule;
 import org.assertj.core.api.Assertions;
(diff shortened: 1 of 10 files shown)


jakubmiarka:
Too quick to review, Travis is unhappy.

src/test/java/unit/uk/gov/ida/verifyserviceprovider/resources/GenerateAuthnRequestResourceTest.java
```diff
@@ -106,16 +114,16 @@ public void samlRequestIsBase64EncodedAuthnRequest() {
...
+                .post(Entity.entity(ImmutableMap.of("bad", "json"), MediaType.APPLICATION_JSON_TYPE));
         assertThat(response.getStatus()).isEqualTo(422);
-        assertThat(response.readEntity(ErrorMessage.class)).isEqualTo(new ErrorMessage(
-            422,
-            "levelOfAssurance may not be null")
-        );
+        ErrorMessage errorMessage = response.readEntity(ErrorMessage.class);
+        assertThat(errorMessage.getCode()).isEqualTo(422);
+        assertThat(errorMessage.getMessage()).startsWith("Unrecognized field \"bad\"");
     }
```
kerrr: Shouldn't we be returning 422 in the "{}" case too? Or that be valid and not throw a 500.

chrisholmes: {} should be valid, but no body should also be valid - I don't think that's being tested.

PR #181, opened 2019-03-05
bau: upgrade to latest dropwizard

chrisholmes:
There's a minor release that fixes a few things and includes a few CVEs. Release notes are [here](https://github.com/dropwizard/dropwizard/blob/v1.3.9/docs/source/about/release-notes.rst).


jakubmiarka:
Can we update the RELASE NOTES as well? It refers to the dropwizard version

PR #193, opened 2019-04-29
BUGFIX for the startup warning in the VSP

j-maynard:
Co-authored-by: Jakub Miarka
<Jakub.Miarka@digital.cabinet-office.gov.uk>


jakubmiarka:
[During start-up VSP is giving a warning due to a use of a generic type](https://trello.com/c/XJXJ5pAz/121-during-start-up-vsp-is-giving-a-warning-due-to-a-use-of-a-generic-type)

PR #195, opened 2019-05-14
vGSUtQgY: Add a new config which allows for overriding hub env values

jakubmiarka:
VSP is now configured using VERIFY_ENVIRONMENT env variable. With options
(COMPLIANCE_TOOL, INTEGRATION, PRODUCTION). However, we want to
run in the staging environment. So we need to override the `ssoLocation` and `metadata URI`
which should be enough to repoint the VSP to staging.

The custom config file is only used in PaaS when connecting to staging.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -136,6 +136,7 @@ distributions {
             from {
                 [
                     'configuration/vsp-with-matching.yml',
+                    'configuration/vsp-with-matching-custom.yml',
                     'configuration/vsp-no-eidas.yml',
                     'configuration/vsp-eidas-without-matching.yml',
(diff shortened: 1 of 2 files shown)


jakubmiarka:
[Add some new test services into staging so we can test rotations](https://trello.com/c/vGSUtQgY/346-add-some-new-test-services-into-staging-so-we-can-test-rotations)

bravegrape:
[Enable Daniel the Manual Spaniel for Verify Team Manual](https://trello.com/c/cQ6eEE0m/441-enable-daniel-the-manual-spaniel-for-verify-team-manual)

PR #197, opened 2019-05-31
BAU: enable Codacy coverage reporting

bjgill:
With GitHub authentication so we avoid being rate-limited


bjgill:
Not sure why travis has started failing. Please hold off on reviewing whilst I investigate...

bjgill:
I still don't understand. Somehow, adding `apply plugin: 'jacoco'` causes `returns500WhenManifestReaderThrowsException` to start failing on Java 8.

That test involves throwing an exception and expecting a 500 response. It looks as if something(?) is causing that exception to not get caught as normal but end up getting propagated to the test...

bjgill:
I've found https://stackoverflow.com/questions/49154559/test-failing-when-using-jacoco, which perhaps suggests that the failure might have something to do with all the reflection going on in the test?

In any case, I've not been able to come up with a workaround. I'm not sure this is something worth spending a huge amount of time on. I'm going to close this and leave for the team supporting the VSP to pick up if they think it worthwhile.

PR #201, opened 2019-06-26
DEUtoBzK Added new production manifest to the VSP

j-maynard:
We need a product version of the VSP in the hub for testing against this change creates a new configuration file for the VSP for production.


brenetic:
Could we have a little description? I don't really have context of the VSP and where the production manifest is going to run etc

PR #202, opened 2019-06-26
PeUHqcqv: constraint to stop eIDAS and MSA config being set together

chrisholmes:
This change adds a constraint to the VSP's config so that the `europeanIdentity` and `msa` configuration can't be set at the same time. Setting this behavior together shouldn't work as the two modes are mutually exclusive. eIDAS support on the VSP is only needed when its acting in a standalone capacity.

### DIFF
--- a/src/main/java/uk/gov/ida/verifyserviceprovider/configuration/VerifyServiceProviderConfiguration.java
+++ b/src/main/java/uk/gov/ida/verifyserviceprovider/configuration/VerifyServiceProviderConfiguration.java
@@ -1,9 +1,11 @@
 package uk.gov.ida.verifyserviceprovider.configuration;
 
-import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import io.dropwizard.Configuration;
+import io.dropwizard.validation.ValidationMethod;
+import org.hibernate.validator.valuehandling.UnwrapValidatedValue;
 import org.joda.time.Duration;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/ida/verifyserviceprovider/configuration/VerifyServiceProviderConfiguration.java
```diff
@@ -98,4 +100,10 @@ public Duration getClockSkew() {
     public Optional<EuropeanIdentityConfiguration> getEuropeanIdentity() {
         return europeanIdentity;
     }
+
+    @ValidationMethod(message = "eIDAS and MSA support cannot be enabled together. The VSP's eIDAS support is only support when it operates without the MSA")
+    @JsonIgnore
+    public boolean isNotMsaAndEidas() {
+        return !(msaMetadata.isPresent() && europeanIdentity.isPresent());
```
jakubmiarka: what if `europeanIdentity` is present but set to `enabled: false`? Then the error message is confusing...

chrisholmes: The configuration shouldn't be there at all if you're doing matching - setting `europeanIdentity.enabled` to `false` won't do anything.

jakubmiarka: in that case I would be inclined to update the error message to be clear, that the configuration should not be present. Otherwise it might confuse people if they somehow have it there but `enabled: false` and getting error message it cannot be enabled (which from their point of view is not)

chrisholmes: Fair enough, I'll change the message to say 'set' rather than 'enabled'

PR #207, opened 2019-07-02
PeUHqcqv: rename classes for new naming convention

chrisholmes:
We decided earlier that we would use the term IdentityX to refer to
objects that previously were called NonMatchingX. When we
would say something `NonMatchingAttribute` we now say
`IdentityAttribute`. We think this better matches the domain model we're
aiming for.

I've also renamed some Dtos that were used for the user account creation flow to add an `AccountCreation` prefix. That way we can use more genenric names for the non-matching code so `NonMatchingAddress` has become `Address`.


j-maynard (dismissed):
🤘🏻

kerrr (dismissed):

PR #214, opened 2019-10-02
EID-1683 rely on AudienceRestrictionValidator from saml-libs

instantiator:
This PR is part 3 of 3 PRs that will form EID-1683. It:

- ...uses the latest update to saml-libs (build 219).
- ...removes `AudienceRestrictionValidator`, `SamlResponseValidationException` from the MSA
- ...uses versions from `saml-libs` instead.
- ...prepares the MSA for new configuration with multiple acceptable entityIds for audience validation.
- ...gracefully falls back to the default behaviour if no additional configuration is found.

For reference, this is very similar to the changes made in the MSA - see: https://github.com/alphagov/verify-matching-service-adapter/pull/140


src/main/java/uk/gov/ida/verifyserviceprovider/configuration/EuropeanIdentityConfiguration.java
```diff
@@ -38,6 +41,13 @@ public EuropeanIdentityConfiguration(@JsonProperty("hubConnectorEntityId") Strin
         this.enabled = enabled;
         this.hubConnectorEntityId = hubConnectorEntityId;
         this.trustStoreConfiguration = trustStore;
+
+        this.acceptableHubConnectorEntityIds =
+            ArrayUtils.isEmpty(acceptableHubConnectorEntityIds)
+                ? new String[] { hubConnectorEntityId }
+                : ArrayUtils.contains(acceptableHubConnectorEntityIds, hubConnectorEntityId)
```
CharlesIC: Would be better to do `unique set` or `unique list` or something like that. You can also use lambda nicely here along the lines of: create new list containing `hubConnectorEntityId `, then add `acceptableHubConnectorEntityIds ` and make a unique set 

Also we don't use `ArrayUtils` - this is a 3rd party library. Can we use `Arrays.asList` or `Arrrays.stream` instead so we stick to using Java built-ins?

Also it would help if `acceptableHubConnectorEntityIds` was a list - using an array is clunky

src/main/java/uk/gov/ida/verifyserviceprovider/configuration/EuropeanIdentityConfiguration.java
```diff
@@ -54,6 +64,11 @@ public String getHubConnectorEntityId() {
                 .orElse(environment.getEidasHubConnectorEntityId());
     }
 
+    public String[] getAcceptableHubConnectorEntityIds() {
+        return Optional.ofNullable(acceptableHubConnectorEntityIds)
```
CharlesIC: Why are there two placed to obtain that value....? This doesn't seem right and is confusing.

src/main/java/uk/gov/ida/verifyserviceprovider/configuration/HubEnvironment.java
```diff
@@ -109,6 +111,14 @@ public String getEidasHubConnectorEntityId(){
         return this.eidasHubConnectorEntityId;
     }
 
+    public String[] getEidasAcceptableHubConnectorEntityIds() {
+        return ArrayUtils.isEmpty(eidasAcceptableHubConnectorEntityIds)
```
CharlesIC: Would be better to do `unique set` or `unique list` or something like that. You can also use lambda/streams nicely here along the lines of: create new list containing hubConnectorEntityId , then add acceptableHubConnectorEntityIds and make a unique set

Also we don't use `ArrayUtils` - this is a 3rd party library. Can we use `Arrays.asList` or `Arrrays.stream` instead so we stick to using Java built-ins?

Also it would help if `acceptableHubConnectorEntityIds` was a list - using an array is clunky

src/main/java/uk/gov/ida/verifyserviceprovider/configuration/EuropeanIdentityConfiguration.java
instantiator: This might be a good question to ask to the original author.

CharlesIC: 🤦‍♂

CharlesIC: Can we make it better?

src/main/java/uk/gov/ida/verifyserviceprovider/configuration/EuropeanIdentityConfiguration.java
```diff
@@ -38,6 +41,13 @@ public EuropeanIdentityConfiguration(@JsonProperty("hubConnectorEntityId") Strin
         this.enabled = enabled;
         this.hubConnectorEntityId = hubConnectorEntityId;
         this.trustStoreConfiguration = trustStore;
+
+        this.acceptableHubConnectorEntityIds =
+            ArrayUtils.isEmpty(acceptableHubConnectorEntityIds)
+                ? new String[] { hubConnectorEntityId }
+                : ArrayUtils.contains(acceptableHubConnectorEntityIds, hubConnectorEntityId)
+                    ? acceptableHubConnectorEntityIds
+                    : (String[])ArrayUtils.add(acceptableHubConnectorEntityIds, hubConnectorEntityId);
```
CharlesIC: Would be better to do `unique set` or `unique list` or something like that. You can also use lambda nicely here along the lines of: create new list containing `hubConnectorEntityId `, then add `acceptableHubConnectorEntityIds ` and make a unique set 

Also we don't use `ArrayUtils` - this is a 3rd party library. Can we use `Arrays.asList` or `Arrrays.stream` instead so we stick to using Java built-ins?

Also it would help if `acceptableHubConnectorEntityIds` was a list - using an array is clunky

src/main/java/uk/gov/ida/verifyserviceprovider/configuration/HubEnvironment.java
```diff
@@ -109,6 +111,14 @@ public String getEidasHubConnectorEntityId(){
...
+            ? new String[] { eidasHubConnectorEntityId }
+            : ArrayUtils.contains(eidasAcceptableHubConnectorEntityIds, eidasHubConnectorEntityId)
+                ? eidasAcceptableHubConnectorEntityIds
+                : (String[])ArrayUtils.add(eidasAcceptableHubConnectorEntityIds, eidasHubConnectorEntityId);
+    }
```
CharlesIC: Would be better to do `unique set` or `unique list` or something like that. You can also use lambda nicely here along the lines of: create new list containing `hubConnectorEntityId `, then add `acceptableHubConnectorEntityIds ` and make a unique set 

Also we don't use `ArrayUtils` - this is a 3rd party library. Can we use `Arrays.asList` or `Arrrays.stream` instead so we stick to using Java built-ins?

Also it would help if `acceptableHubConnectorEntityIds` was a list - using an array is clunky

src/main/java/uk/gov/ida/verifyserviceprovider/configuration/EuropeanIdentityConfiguration.java
instantiator: I've added `acceptableHubConnectorEntityIds` wherever `hubConnectorEntityId` is defined.

`hubConnectorEntityId` is defined in `EuropeanIdentityConfiguration` (apparently constructed from json), and `HubEnvironment` (an enum with a lot of hard coded params). There's very likely a good reason why it's the way it is. It may take some time to figure out. Happy to look into it.

src/main/java/uk/gov/ida/verifyserviceprovider/configuration/EuropeanIdentityConfiguration.java
```diff
@@ -38,6 +44,11 @@ public EuropeanIdentityConfiguration(@JsonProperty("hubConnectorEntityId") Strin
         this.enabled = enabled;
         this.hubConnectorEntityId = hubConnectorEntityId;
         this.trustStoreConfiguration = trustStore;
+
+        Set<String> hubConnectorEntityIds = new LinkedHashSet<>();
+        Optional.ofNullable(acceptableHubConnectorEntityIds).ifPresent(hubConnectorEntityIds::addAll);
+        Optional.ofNullable(hubConnectorEntityId).ifPresent(hubConnectorEntityIds::add);
+        this.acceptableHubConnectorEntityIds = new LinkedList<>(hubConnectorEntityIds);
```
Wynndow: Again, out of interest, why the LinkedList rather than a plain old vanilla List?

src/main/java/uk/gov/ida/verifyserviceprovider/configuration/EuropeanIdentityConfiguration.java
```diff
@@ -38,6 +44,11 @@ public EuropeanIdentityConfiguration(@JsonProperty("hubConnectorEntityId") Strin
         this.enabled = enabled;
         this.hubConnectorEntityId = hubConnectorEntityId;
         this.trustStoreConfiguration = trustStore;
+
+        Set<String> hubConnectorEntityIds = new LinkedHashSet<>();
```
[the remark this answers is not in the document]
instantiator: First implementation of HashSet I encountered.

src/main/java/uk/gov/ida/verifyserviceprovider/configuration/EuropeanIdentityConfiguration.java
instantiator: I think the alternative is plain old vanilla ArrayList (as List is an interface). I think IntelliJ autocompletes/offers LinkedList preferentially (or maybe I imagined that). Again, no strong feelings, but it's been my go-to List implementation for a while.

Wynndow: Cool! No complaints from me, just interested.

src/test/java/unit/uk/gov/ida/verifyserviceprovider/services/EidasAssertionTranslatorTest.java
```diff
@@ -98,12 +98,13 @@ public void setUp() {
...
             signatureValidatorFactory,
             HUB_CONNECTOR_ENTITY_ID,
+            singletonList(HUB_CONNECTOR_ENTITY_ID),
             userIdHashFactory);
         doNothing().when(instantValidator).validate(any(), any());
         doNothing().when(subjectValidator).validate(any(), any());
         doNothing().when(conditionsValidator).validate(any(), any());
         doNothing().when(levelOfAssuranceValidator).validate(any(), any());
-        when(metadataResolverRepository.getResolverEntityIds()).thenReturn(asList(STUB_COUNTRY_ONE));
+        when(metadataResolverRepository.getResolverEntityIds()).thenReturn(singletonList(STUB_COUNTRY_ONE));
```
Wynndow: What's the benefit of a singletonList here?

instantiator: IntelliJ highlighted it and offered singletonList. I've not yet explored singletonList, but Array.asList doesn't actually make a List. It makes something that implements List but throws exceptions if you try to add/remove items - which I do have feelings about.

JohnWattsGDS: Off the top of my head, I think it creates an immutable list (which is why it bogs out when you try to add or remove from the list).

src/main/java/uk/gov/ida/verifyserviceprovider/configuration/ConfigurationConstants.java
```diff
@@ -38,16 +35,41 @@
     String PRODUCTION_METADATASOURCE_URI = PRODUCTION_HOST + METADATASOURCE_PATH;
     String PRODUCTION_HUBCONNECTOR_ENTITY_ID = PRODUCTION_HOST + HUBCONNECTOR_ENTITY_ID_PATH;
 
+    String[] PRODUCTION_ACCEPTABLE_HUBCONNECTOR_ENTITY_IDS = new String[] {
+        "https://www.signin.service.gov.uk/SAML2/metadata/connector",               // (AWS)
+        "https://connector-node.london.verify.govsvc.uk/ConnectorMetadata",         // (GSP)
+        "https://eidas.signin.service.gov.uk/ConnectorMetadata",                    // (potential new #1)
+        "https://connector.eidas.signin.service.gov.uk/ConnectorMetadata",          // (potential new #2)
+        "https://connector-node.eidas.signin.service.gov.uk/ConnectorMetadata",     // (potential new #3)
```
idrop: After seeing this on the [MSA PR](https://github.com/alphagov/verify-matching-service-adapter/pull/142) I'm wondering if we should put these in `saml-libs` (`IdaConstants` for instance).

instantiator: Have also discussed with Karol and I'll add it to the related tech debt story for this so we can make a decision later (once the entity ids are finalised, and we've got this thing out the door). As you mentioned, `saml-libs` is good to share things, but risks becoming an overgrown do-everything-be-everything, and I wonder if there's a better/more appropriate way to share fixed config values like this.

PR #216, opened 2019-10-09
EID-1699 update VSP release notes for entityId and unsigned assertions work

instantiator:
A small addition to `RELEASE_NOTES.md`


src/test/resources/verify-service-provider-with-msa-and-eidas.yml
```diff
@@ -57,7 +57,6 @@ msaMetadata:
 
 europeanIdentity:
   enabled: ${EUROPEAN_IDENTITY_ENABLED}
-  hubConnectorEntityId: ${HUB_CONNECTOR_ENTITY_ID}
```
CharlesIC: This file is used to test the config logic so actually let's not touch it

PR #219, opened 2019-10-11
EID-1607 Handle unsigned assertions

Wynndow:
If you want to run the VSP locally so you can debug it, there are a couple of branches which will help. These are by no means meant to be a good way to run it, it's just how I hacked it together when developing this PR. Use these branches locally and then run `./hubstartup.sh` in `ida-hub-acceptance-tests`
https://github.com/alphagov/ida-hub-acceptance-tests/tree/test-rp-vsp
https://github.com/alphagov/verify-test-rp/tree/non-matching

You'll also need to cherry pick the last commit from this branch of the VSP, which adds a local config: https://github.com/alphagov/verify-service-provider/tree/eid-1607

### EID-1607 Bump saml-lib to build 221
It includes a few updates including making it easier to create a
decrypter for unsigned assertions.

### EID-1607 Update ResponseService to use `UnsignedAssertionsResponseHandler`
The ResponseService needs to be able to determine if the incoming
Response has unsigned assertions wrapped up inside it. We do this with a
private helper method which interrogates the names of the attribute
values within the assertion, and returns false if any match the Constant
defined for this purpose, which will have been set by the hub.

If it determines that the assertion _does_ contain eIDAS unsigned
assertions it uses the new UnsignedAssertionsResponseHandler to first
validate the original eIDAS SAML response, and then if valid, decrypt
the assertions.

These assertions are then passed to the assertion translator instead of
the original wrapper assertion.

### EID-1607 Create UnsignedAssertionsResponseHandler …
This class will get called by the response service to handle the wrapper
Response that contains the original unsigned response from the country,
and the encryption keys required to decrypt the assertions it contains.

It uses the `eidasValidatorFactory` from saml-lib to validate the
original response signature.

The other interesting method `decryptAssertion` will iterate through all
the keys supplied in the wrapper Response assertion and attempt to
decrypt the assertions in the validated country response. Most of the
time we only expect to receive one key, but we do pass a list so it's
worth trying all of them.

We also pass into this a `EidasEncryptionAlgorithmValidator` which is a
new class. The eIDAS spec defines different encryption algorithms to
that which the hub uses, so we need to make sure we're validating the
country response against the right things.

### EID-1607 Add `EidasUnsignedAssertionTranslator` …
In an eIDAS enabled non-matching set up, the `assertionTranslator` used
by the `ResponseService` is a `ClassifyingAssertionTranslator`.

This translator is configured with different translators and will
determine which to use depending on the assertions given to it.

This extends it to include the new `EidasUnsignedAssertionTranslator`
which will deal with unsigned assertions.

The only difference in dealing with signed and unsigned assertions is
the validation of the signature. As such, the vast majority of the logic
in `EidasAssertionTranslator` has been moved to an abstract base class
which is extended by the new `EidasAssertionTranslator` as well as the
new one. They both just implement the abstract `validateSignature`
method.

### EID 1607 Create UnsignedAssertionsResponseHandler in VerifyServiceProviderFactory
The new `UnsignedAssertionsResponseHandler` only needs creating if the
VSP is configured to run as an eIDAS enabled non-matching VSP. As such,
we create the new response handler in the relevant method of the
`VerifyServiceProviderFactory`.

We also create the `EidasUnsignedAssertionTranslator` here which is
required by the `ClassifyingAssertionTranslator`.

### EID-1607 Create `createEidasUnsignedAssertionService` in ResponseFactory
This allows us to create a new `EidasUnsignedAssertionTranslator`, and
is called from the `VerifyServiceProviderFactory`. It's a similar
pattern to how the `EidasAssertionTranslator` is created.

Also tidy up some tests.

### EID-1607 Add an acceptance test for unsigned assertions
This test creates an eIDAS Response with an unsigned encrypted
assertion. It then converts that Response into a string and stuffs it
into a hub Response.

The hub Response has two attributes in its assertion. One is the
original eIDAS response, the other is the encryption keys used in that
response, which have been extracted and encrypted for the RP.

The encryption algorithm used in the eIDAS response has been changed to
be one of the eIDAS spec algorithms.

### EID-1607 Validate hub assertion signature
If we're dealing with unsigned assertions, the assertion containing the
original eIDAS SAML and encrypted assertion keys will be signed by the
hub. We should validate that signature.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -34,7 +34,7 @@ project.ext {
     version_number = '2.0.0'
     openSamlVersion = '3.4.3'
     verifyCommonUtils = '2.0.0-337'
-    samlLibVersion = "$openSamlVersion-221"
+    samlLibVersion = "$openSamlVersion-224"
     dropwizardVersion = '1.3.9'
     jaxbapiVersion = '2.2.9'
 }
--- a/build.gradle
+++ b/build.gradle
@@ -34,7 +34,7 @@ project.ext {
     version_number = '2.0.0'
     openSamlVersion = '3.4.3'
     verifyCommonUtils = '2.0.0-337'
-    samlLibVersion = "$openSamlVersion-219"
+    samlLibVersion = "$openSamlVersion-221"
     dropwizardVersion = '1.3.9'
     jaxbapiVersion = '2.2.9'
 }
--- a/src/main/java/uk/gov/ida/verifyserviceprovider/exceptions/MissingUnsignedAssertionsHandlerException.java
+++ b/src/main/java/uk/gov/ida/verifyserviceprovider/exceptions/MissingUnsignedAssertionsHandlerException.java
@@ -1,0 +1,7 @@
+package uk.gov.ida.verifyserviceprovider.exceptions;
+
+public class MissingUnsignedAssertionsHandlerException extends RuntimeException {
+    public MissingUnsignedAssertionsHandlerException() {
+        super("UnsignedAssertionsHandler missing: UnsignedAssertionsHandler can not be null in eIDAS enabled VSP");
+    }
+}
--- a/src/main/java/uk/gov/ida/verifyserviceprovider/factories/saml/ResponseFactory.java
+++ b/src/main/java/uk/gov/ida/verifyserviceprovider/factories/saml/ResponseFactory.java
@@ -98,7 +98,7 @@ public class ResponseFactory {
             new SamlResponseSignatureValidator(new SamlMessageSignatureValidator(metadataBackedSignatureValidator)),
             new InstantValidator(dateTimeComparator),
             new MatchingResponderCodeTranslator(),
-            Optional.empty()
+            null
         );
     }
 
@@ -106,7 +106,7 @@ public class ResponseFactory {
             ExplicitKeySignatureTrustEngine hubSignatureTrustEngine,
             AssertionTranslator nonMatchingAssertionTranslator,
             DateTimeComparator dateTimeComparator,
-            Optional<UnsignedAssertionsResponseHandler> unsignedAssertionsResponseHandler
+            UnsignedAssertionsResponseHandler unsignedAssertionsResponseHandler
     ) {
         AssertionDecrypter assertionDecrypter = createAssertionDecrypter();
         MetadataBackedSignatureValidator metadataBackedSignatureValidator = createMetadataBackedSignatureValidator(hubSignatureTrustEngine);
--- a/src/main/java/uk/gov/ida/verifyserviceprovider/factories/saml/ResponseFactory.java
+++ b/src/main/java/uk/gov/ida/verifyserviceprovider/factories/saml/ResponseFactory.java
@@ -38,6 +38,7 @@ import uk.gov.ida.verifyserviceprovider.services.VerifyAssertionTranslator;
 import uk.gov.ida.verifyserviceprovider.utils.DateTimeComparator;
 import uk.gov.ida.verifyserviceprovider.validators.AssertionValidator;
 import uk.gov.ida.verifyserviceprovider.validators.ConditionsValidator;
+import uk.gov.ida.verifyserviceprovider.validators.EidasAssertionTranslatorValidatorContainer;
 import uk.gov.ida.verifyserviceprovider.validators.InstantValidator;
 import uk.gov.ida.verifyserviceprovider.validators.LevelOfAssuranceValidator;
 import uk.gov.ida.verifyserviceprovider.validators.ResponseSizeValidator;
@@ -46,7 +47,6 @@ import uk.gov.ida.verifyserviceprovider.validators.TimeRestrictionValidator;
 
 import java.security.KeyPair;
 import java.util.List;
-import java.util.Optional;
 
 public class ResponseFactory {
 
@@ -168,16 +168,10 @@ public class ResponseFactory {
             EuropeanIdentityConfiguration europeanIdentityConfiguration,
             String hashingEntityId
     ) {
-        TimeRestrictionValidator timeRestrictionValidator = new TimeRestrictionValidator(dateTimeComparator);
-        AudienceRestrictionValidator audienceRestrictionValidator = new AudienceRestrictionValidator();
-
         return new EidasAssertionTranslator(
-                new SubjectValidator(timeRestrictionValidator),
+                getEidasAssertionValidatorContainer(dateTimeComparator),
                 new EidasMatchingDatasetUnmarshaller(),
                 new MatchingDatasetToNonMatchingAttributesMapper(),
-                new InstantValidator(dateTimeComparator),
-                new ConditionsValidator(timeRestrictionValidator, audienceRestrictionValidator),
-                new LevelOfAssuranceValidator(),
                 eidasMetadataResolverRepository,
                 new SignatureValidatorFactory(),
                 europeanIdentityConfiguration.getAllAcceptableHubConnectorEntityIds(),
@@ -191,16 +185,10 @@ public class ResponseFactory {
             EuropeanIdentityConfiguration europeanIdentityConfiguration,
             String hashingEntityId
     ) {
-        TimeRestrictionValidator timeRestrictionValidator = new TimeRestrictionValidator(dateTimeComparator);
-        AudienceRestrictionValidator audienceRestrictionValidator = new AudienceRestrictionValidator();
-
         return new EidasUnsignedAssertionTranslator(
-                new SubjectValidator(timeRestrictionValidator),
+                getEidasAssertionValidatorContainer(dateTimeComparator),
                 new EidasMatchingDatasetUnmarshaller(),
                 new MatchingDatasetToNonMatchingAttributesMapper(),
-                new InstantValidator(dateTimeComparator),
-                new ConditionsValidator(timeRestrictionValidator, audienceRestrictionValidator),
-                new LevelOfAssuranceValidator(),
                 eidasMetadataResolverRepository,
                 europeanIdentityConfiguration.getAllAcceptableHubConnectorEntityIds(),
                 new UserIdHashFactory(hashingEntityId)
@@ -213,5 +201,15 @@ public class ResponseFactory {
 
     private IdaKeyStore createEncryptionKeyStore() {
         return new IdaKeyStore(null, encryptionKeyPairs);
+    }
+
+    private EidasAssertionTranslatorValidatorContainer getEidasAssertionValidatorContainer(DateTimeComparator dateTimeComparator) {
+        TimeRestrictionValidator timeRestrictionValidator = new TimeRestrictionValidator(dateTimeComparator);
+        return new EidasAssertionTranslatorValidatorContainer(
+                new SubjectValidator(timeRestrictionValidator),
+                new InstantValidator(dateTimeComparator),
+                new ConditionsValidator(timeRestrictionValidator, new AudienceRestrictionValidator()),
+                new LevelOfAssuranceValidator()
+        );
     }
 }
--- a/src/main/java/uk/gov/ida/verifyserviceprovider/factories/saml/ResponseFactory.java
+++ b/src/main/java/uk/gov/ida/verifyserviceprovider/factories/saml/ResponseFactory.java
@@ -123,7 +123,7 @@ public class ResponseFactory {
         );
     }
 
-    public MatchingAssertionTranslator createMsaAssertionService(
+    public MatchingAssertionTranslator createMsaAssertionTranslator(
             ExplicitKeySignatureTrustEngine signatureTrustEngine,
             SignatureValidatorFactory signatureValidatorFactory,
             DateTimeComparator dateTimeComparator
@@ -144,9 +144,9 @@ public class ResponseFactory {
         );
     }
 
-    public VerifyAssertionTranslator createVerifyIdpAssertionService(SamlAssertionsSignatureValidator hubSignatureValidator,
-                                                                     DateTimeComparator dateTimeComparator,
-                                                                     String hashingEntityId) {
+    public VerifyAssertionTranslator createVerifyIdpAssertionTranslator(SamlAssertionsSignatureValidator hubSignatureValidator,
+                                                                        DateTimeComparator dateTimeComparator,
+                                                                        String hashingEntityId) {
 
         TimeRestrictionValidator timeRestrictionValidator = new TimeRestrictionValidator(dateTimeComparator);
 
@@ -162,7 +162,7 @@ public class ResponseFactory {
             );
     }
 
-    public EidasAssertionTranslator createEidasAssertionService(
+    public EidasAssertionTranslator createEidasAssertionTranslator(
             DateTimeComparator dateTimeComparator,
             EidasMetadataResolverRepository eidasMetadataResolverRepository,
             EuropeanIdentityConfiguration europeanIdentityConfiguration,
@@ -185,7 +185,7 @@ public class ResponseFactory {
         );
     }
 
-    public EidasUnsignedAssertionTranslator createEidasUnsignedAssertionService(
+    public EidasUnsignedAssertionTranslator createEidasUnsignedAssertionTranslator(
             DateTimeComparator dateTimeComparator,
             EidasMetadataResolverRepository eidasMetadataResolverRepository,
             EuropeanIdentityConfiguration europeanIdentityConfiguration,
--- a/src/main/java/uk/gov/ida/verifyserviceprovider/factories/saml/ResponseFactory.java
+++ b/src/main/java/uk/gov/ida/verifyserviceprovider/factories/saml/ResponseFactory.java
@@ -144,15 +144,14 @@ public class ResponseFactory {
         );
     }
 
-    public VerifyAssertionTranslator createVerifyIdpAssertionService(ExplicitKeySignatureTrustEngine signatureTrustEngine,
-                                                                     SignatureValidatorFactory signatureValidatorFactory,
+    public VerifyAssertionTranslator createVerifyIdpAssertionService(SamlAssertionsSignatureValidator hubSignatureValidator,
                                                                      DateTimeComparator dateTimeComparator,
                                                                      String hashingEntityId) {
 
         TimeRestrictionValidator timeRestrictionValidator = new TimeRestrictionValidator(dateTimeComparator);
 
         return new VerifyAssertionTranslator(
-                signatureValidatorFactory.getSignatureValidator(signatureTrustEngine),
+                hubSignatureValidator,
                 new SubjectValidator(timeRestrictionValidator),
                 new AssertionAttributeStatementValidator(),
                 new VerifyMatchingDatasetUnmarshaller(new AddressFactory()),
--- a/src/main/java/uk/gov/ida/verifyserviceprovider/factories/saml/ResponseFactory.java
+++ b/src/main/java/uk/gov/ida/verifyserviceprovider/factories/saml/ResponseFactory.java
@@ -16,7 +16,6 @@ import uk.gov.ida.saml.deserializers.validators.NotNullSamlStringValidator;
 import uk.gov.ida.saml.metadata.EidasMetadataResolverRepository;
 import uk.gov.ida.saml.security.AssertionDecrypter;
 import uk.gov.ida.saml.security.DecrypterFactory;
-import uk.gov.ida.saml.security.EidasValidatorFactory;
 import uk.gov.ida.saml.security.IdaKeyStore;
 import uk.gov.ida.saml.security.IdaKeyStoreCredentialRetriever;
 import uk.gov.ida.saml.security.MetadataBackedSignatureValidator;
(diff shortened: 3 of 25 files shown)


src/main/java/uk/gov/ida/verifyserviceprovider/services/EidasUnsignedAssertionTranslator.java
```diff
@@ -0,0 +1,44 @@
...
+import uk.gov.ida.verifyserviceprovider.factories.saml.UserIdHashFactory;
+import uk.gov.ida.verifyserviceprovider.mappers.MatchingDatasetToNonMatchingAttributesMapper;
+import uk.gov.ida.verifyserviceprovider.validators.ConditionsValidator;
+import uk.gov.ida.verifyserviceprovider.validators.InstantValidator;
+import uk.gov.ida.verifyserviceprovider.validators.LevelOfAssuranceValidator;
+import uk.gov.ida.verifyserviceprovider.validators.SubjectValidator;
+
+import java.util.List;
+
+public class EidasUnsignedAssertionTranslator extends BaseEidasAssertionTranslator {
```
JohnWattsGDS: In no way a blocking comment, but maybe in future consider composition for this sort of thing - have one EidasAssertionTranslator and two classes that implement the signature validation (interface) which are injected at creation (or get all modern and funky and use lambdas although that wouldn't give the testability that separate classes would).  Clearly one implementation would still be a no-op implementation.

src/test/java/unit/uk/gov/ida/verifyserviceprovider/saml/ResponseFactoryTest.java
```diff
@@ -20,26 +21,25 @@
...
-    public void shouldNotAllowNullSamlResponse() {
+    public void createStringToResponseTransformerShouldNotAllowNullSamlResponse() {
         expectedException.expect(SamlTransformationErrorException.class);
         expectedException.expectMessage("SAML Validation Specification: Missing SAML message.");
         stringToResponseTransformer.apply(null);
     }
 
     @Test
-    public void shouldNotAllowNotBase64EncodedSamlResponse() {
+    public void createStringToResponseTransformerShouldNotAllowNotBase64EncodedSamlResponse() {
```
galund: Too many "nots" :-D  -- how about `createStringToResponseTransformerMustContainBase64EncodedSamlResponse`?

src/main/java/uk/gov/ida/verifyserviceprovider/factories/saml/ResponseFactory.java
```diff
@@ -182,7 +182,29 @@ public EidasAssertionTranslator createEidasAssertionService(
                 new SignatureValidatorFactory(),
                 europeanIdentityConfiguration.getAllAcceptableHubConnectorEntityIds(),
                 new UserIdHashFactory(hashingEntityId)
-                );
+        );
+    }
+
+    public EidasUnsignedAssertionTranslator createEidasUnsignedAssertionService(
```
galund: I see this is already the convention, but I'm finding it confusing that functions that provide a 'Translator' are named as if they are providing a 'Service'. Is a 'Translator' a service, I guess so, but it seems strange that we flip how we call it. There's also some inconsistency here about how we are using a factory instead of the DI that we do elsewhere, but whatevs :-)

src/main/java/uk/gov/ida/verifyserviceprovider/services/ResponseService.java
```diff
@@ -72,16 +70,30 @@ public TranslatedResponseBody convertTranslatedResponseBody(
                 return responderCodeTranslator.translateResponderCode(statusCode);
             case StatusCode.SUCCESS:
                 List<Assertion> assertions = assertionDecrypter.decryptAssertions(validatedResponse);
+                if (assertionsContainEidasUnsignedAssertionsResponse(assertions)) {
+                    UnsignedAssertionsResponseHandler handler = unsignedAssertionsResponseHandler.get();
```
instantiator: _In no way a blocker, and probably just a style choice..._ The `unsignedAssertionsResponseHandler` is `Optional`, but we straight-up `get()` it. It feels like maybe it's not optional and always present? (Or always present when we need it?)

I guess so long as we throw an exception if it's not there when we need it, then that makes sense - but we can achieve that without `Optional`.

src/main/java/uk/gov/ida/verifyserviceprovider/services/ResponseService.java
```diff
@@ -72,16 +70,30 @@ public TranslatedResponseBody convertTranslatedResponseBody(
...
+
+                    ValidatedResponse validatedCountryResponse = handler.getValidatedResponse(assertions.get(0), expectedInResponseTo);
```
instantiator: _Personal style choice (again!)..._ when putting a constant like `0` into code, it feels awkward not to explain the assumption behind it (ie. "in this case there can only be 1 assertion and it should contain...") I appreciate we don't write many comments though - so I suspect it's ok not to do this.

src/main/java/uk/gov/ida/verifyserviceprovider/services/ResponseService.java
```diff
@@ -72,16 +70,30 @@ public TranslatedResponseBody convertTranslatedResponseBody(
...
         }
     }
 
-    private ValidatedResponse getValidatedResponse (Response response) {
-        String issuer = response.getIssuer().getValue();
-        return eidasValidatorFactory.isPresent() && !issuer.equals(hubEntityId) ?
-                eidasValidatorFactory.get().getValidatedResponse(response) :
-                responseSignatureValidator.validate(response, SPSSODescriptor.DEFAULT_ELEMENT_NAME);
+    private boolean assertionsContainEidasUnsignedAssertionsResponse(List<Assertion> assertions) {
+        if (assertions == null || assertions.size() != 1) { return false; }
```
instantiator: In fact, perhaps this method could state its assumptions. (If we were to comment on them.)

idrop:
👍The logic of determining the unsigned assertions path, decrypting the eidas saml response, validating the sig on that response, then validating the assertions (except the signing) is functionally equivalent to the [work done for the MSA](https://govukverify.atlassian.net/browse/EID-1604).
LGTM, outside the Codacy issues.

src/test/java/unit/uk/gov/ida/verifyserviceprovider/services/UnsignedAssertionResponseHandlerTest.java
```diff
@@ -0,0 +1,244 @@
...
+import static uk.gov.ida.saml.core.test.builders.AssertionBuilder.anAssertion;
+import static uk.gov.ida.saml.core.test.builders.AssertionBuilder.anEidasAssertion;
+import static uk.gov.ida.saml.core.test.builders.AttributeStatementBuilder.anAttributeStatement;
+import static uk.gov.ida.saml.core.test.builders.ResponseBuilder.aResponse;
+
+@RunWith(MockitoJUnitRunner.class)
+public class UnsignedAssertionResponseHandlerTest {
+
+    @Mock
+    EidasValidatorFactory eidasValidatorFactory;
```
idrop:
```suggestion
    private EidasValidatorFactory eidasValidatorFactory;
```

src/test/java/unit/uk/gov/ida/verifyserviceprovider/services/UnsignedAssertionResponseHandlerTest.java
```diff
@@ -0,0 +1,244 @@
...
+
+    @Mock
+    StringToOpenSamlObjectTransformer<Response> stringToResponseTransformer;
```
idrop:
```suggestion
    private StringToOpenSamlObjectTransformer<Response> stringToResponseTransformer;
```

src/test/java/unit/uk/gov/ida/verifyserviceprovider/services/UnsignedAssertionResponseHandlerTest.java
```diff
@@ -0,0 +1,244 @@
...
+
+    @Mock
+    InstantValidator instantValidator;
```
idrop:
```suggestion
    private InstantValidator instantValidator;
```

src/test/java/unit/uk/gov/ida/verifyserviceprovider/services/UnsignedAssertionResponseHandlerTest.java
```diff
@@ -0,0 +1,244 @@
...
+
+    @Mock
+    SecretKeyDecryptorFactory secretKeyDecryptorFactory;
```
idrop:
```suggestion
    private SecretKeyDecryptorFactory secretKeyDecryptorFactory;
```

src/test/java/unit/uk/gov/ida/verifyserviceprovider/services/UnsignedAssertionResponseHandlerTest.java
```diff
@@ -0,0 +1,244 @@
...
+
+    @Mock
+    Decrypter decrypter;
```
idrop:
```suggestion
    private Decrypter decrypter;
```

src/test/java/unit/uk/gov/ida/verifyserviceprovider/services/UnsignedAssertionResponseHandlerTest.java
```diff
@@ -0,0 +1,244 @@
...
+
+    @Mock
+    SamlAssertionsSignatureValidator hubAssertionSignatureValidator;
```
idrop:
```suggestion
    private SamlAssertionsSignatureValidator hubAssertionSignatureValidator;
```

src/main/java/uk/gov/ida/verifyserviceprovider/services/EidasUnsignedAssertionTranslator.java
Wynndow: Interesting. Might be cleaner, I'll think about it next time.

src/test/java/unit/uk/gov/ida/verifyserviceprovider/saml/ResponseFactoryTest.java
Wynndow: Nice...

src/main/java/uk/gov/ida/verifyserviceprovider/factories/saml/ResponseFactory.java
Wynndow: Yeah, you're totally right. I'm going to change them to be 'Translators' instead of 'Services'.

As for the DI, we don't use Guice to inject stuff for us in the VSP like we do in hub, msa etc. Is that what you mean, or is it an inconsistency in how we do DI in just the VSP?

src/main/java/uk/gov/ida/verifyserviceprovider/services/ResponseService.java
Wynndow: Yeah, it does feel a bit weird. It's an Optional as this class can be instantiated in a couple of ways, depending on how the config is set up (eIDAS enabled or not). The conditional on line 73 means we're only ever running this line of code if we're in an eIDAS enabled VSP, so the `get()` _should_ be ok. But you're right, I should probably throw an Exception if it's not present. Will add it.

src/main/java/uk/gov/ida/verifyserviceprovider/services/ResponseService.java
Wynndow: I like that, I'm going to do it.

build.gradle
```diff
@@ -34,7 +34,7 @@ project.ext {
     version_number = '2.0.0'
     openSamlVersion = '3.4.3'
     verifyCommonUtils = '2.0.0-337'
-    samlLibVersion = "$openSamlVersion-219"
+    samlLibVersion = "$openSamlVersion-221"
```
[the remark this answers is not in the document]
Wynndow: Have bumped

src/main/java/uk/gov/ida/verifyserviceprovider/services/ResponseService.java
Wynndow: Did the other thing, so hopefully it's clearer now.

src/test/java/unit/uk/gov/ida/verifyserviceprovider/services/UnsignedAssertionResponseHandlerTest.java
Wynndow: Have added these all in one commit (well, smooshed into an existing commit to be accurate.)

src/main/java/uk/gov/ida/verifyserviceprovider/services/UnsignedAssertionsResponseHandler.java
```diff
@@ -0,0 +1,129 @@
...
+        this.instantValidator = instantValidator;
+        this.secretKeyDecryptorFactory = secretKeyDecryptorFactory;
+        this.encryptionAlgorithmValidator = encryptionAlgorithmValidator;
+    }
+
+    public ValidatedResponse getValidatedResponse(
+            Assertion hubResponseAssertion,
+            String expectedInResponseTo
+    ) {
+        ValidatedResponse validatedResponse = eidasValidatorFactory.getValidatedResponse(
```
instantiator: It's not a part of this review, as you didn't write it, but `eidasValidatorFactory.getValidatedResponse` has quite a few possible interpretations in my head:

- validate the response (best guess)
- call out to another service to validate the response (similar, but more misleading)
- extract a validated response from inside the passed in parameter
- represent the passed in parameter as validated (ie. assumption is it's already validated, so just cast/copy/box it into a ValidatedResponse, a method you might call `asValidatedResponse`)

I took a look in `saml-lib` and `getValidatedResponse` seems to be doing a validation on the signature(s) in the response using a `SamlResponseSignatureValidator`. My instinct would be to rename that method to `validateResponseSignature`, as it's an action not a getter.

I know it's not for now, and for some reason my slight confusion has led to a small essay, but hopefully it helps to make clear what I mean when I say the context is missing from some things - which means it takes more time to understand what they're doing, and is difficult to remember.

No action for this review, but perhaps we can track a list of things we could do to make things a bit clearer for future developers. (I'm a little worried too that as it's in `saml-lib`, changing the name of the method will lead to issues for a number of other projects.)

Wynndow: It could definitely be named better, but I think you're right that it's used across different projects and so would be hard to change.

src/main/java/uk/gov/ida/verifyserviceprovider/services/UnsignedAssertionsResponseHandler.java
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        instantValidator.validate(validatedResponse.getIssueInstant(), "Response IssueInstant");
+
+        return validatedResponse;
+    }
+
+    public List<Assertion> decryptAssertion(
+            ValidatedResponse validatedResponse,
+            Assertion hubResponseAssertion
```
instantiator: I might have called this parameter `hubResponseAssertionWithEncryptedDecryptionKeys` or something similar. If I understand correctly, it contains a number of keys that can be used to decrypt the assertions.

src/test/java/unit/uk/gov/ida/verifyserviceprovider/services/UnsignedAssertionResponseHandlerTest.java
```diff
@@ -0,0 +1,233 @@
...
+
+    private EncryptionAlgorithmValidator getEncryptionAlgorithmValidator() {
+        return new EncryptionAlgorithmValidator(
+                ImmutableSet.of(
+                        EncryptionConstants.ALGO_ID_BLOCKCIPHER_AES128
+                ),
+                ImmutableSet.of(
+                        EncryptionConstants.ALGO_ID_KEYTRANSPORT_RSAOAEP
+                )
+        );
```
instantiator: Rather than an `EidasEncryptionAlgorithmValidator`? I guess this is more restrictive - so perhaps a better test?

src/main/java/uk/gov/ida/verifyserviceprovider/services/UnsignedAssertionsResponseHandler.java
Wynndow: So this is the assertion from the hub response, which contains the original country saml response, as well as the encrypted decryption keys. Could be a bit of a mouthful to get it all in.

src/main/java/uk/gov/ida/verifyserviceprovider/services/UnsignedAssertionsResponseHandler.java
```diff
@@ -0,0 +1,129 @@
...
+        }
+    }
+
+    private String getCountryResponseStringFromAssertion(Assertion hubResponseAssertion) {
+        List<Attribute> attributes = hubResponseAssertion.getAttributeStatements().get(0).getAttributes();
+        CountrySamlResponse countrySamlResponse = (CountrySamlResponse) attributes.get(0).getAttributeValues().get(0);
+        return countrySamlResponse.getValue();
+    }
+
+    private List<String> getEncryptedAssertionKeysFromAssertion(Assertion hubResponseAssertion) {
```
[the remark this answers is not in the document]
Wynndow: The keys *are* encrypted - for the service. The encrypted keys are extracted from the hub response assertion here, and then used to create an assertion decrypter (in `getAssertionDecrypter(String key)`). That method passes the encrypted key(s) to `secretKeyDecryptorFactory.createDecrypter(key)`, which decrypts the key, to be used in an `AssertionDecrypter`.

src/main/java/uk/gov/ida/verifyserviceprovider/services/BaseEidasAssertionTranslator.java
```diff
@@ -0,0 +1,111 @@
...
+import uk.gov.ida.verifyserviceprovider.validators.InstantValidator;
+import uk.gov.ida.verifyserviceprovider.validators.LevelOfAssuranceValidator;
+import uk.gov.ida.verifyserviceprovider.validators.SubjectValidator;
+
+import java.util.List;
+import java.util.Optional;
+
+import static uk.gov.ida.verifyserviceprovider.dto.NonMatchingScenario.IDENTITY_VERIFIED;
+
+public abstract class BaseEidasAssertionTranslator extends IdentityAssertionTranslator {
```
instantiator: 👍 this is a great pattern

src/main/java/uk/gov/ida/verifyserviceprovider/services/EidasAssertionTranslator.java
```diff
@@ -19,22 +14,10 @@
...
-    private final InstantValidator instantValidator;
-    private final ConditionsValidator conditionsValidator;
-    private final LevelOfAssuranceValidator levelOfAssuranceValidator;
-    private final EidasMetadataResolverRepository metadataResolverRepository;
-    private final SignatureValidatorFactory signatureValidatorFactory;
-    private final List<String> acceptableHubConnectorEntityIds;
-    private UserIdHashFactory userIdHashFactory;
-    private final AuthnContextFactory authnContextFactory = new AuthnContextFactory();
 
+public class EidasAssertionTranslator extends BaseEidasAssertionTranslator {
```
instantiator: ... and this is so much clearer about what it does as a result of the abstract base class.

src/main/java/uk/gov/ida/verifyserviceprovider/services/EidasUnsignedAssertionTranslator.java
```diff
@@ -0,0 +1,44 @@
...
+                metadataResolverRepository,
+                null,
+                acceptableHubConnectorEntityIds,
+                userIdHashFactory
+        );
+    }
+
+    @Override
+    void validateSignature(Assertion assertion, String issuerEntityId) {
+        // No need to validate the signature of an unsigned assertion.
```
instantiator: 👍

src/test/java/unit/uk/gov/ida/verifyserviceprovider/services/UnsignedAssertionResponseHandlerTest.java
Wynndow: The crypto setup that the tests use, use this (`ALGO_ID_BLOCKCIPHER_AES128`) encryption algorithm. This is different to the algorithms specified in `EidasEncryptionAlgorithmValidator` which is why the tests use the validator being returned here.

We test that the handler will validate assertions encrypted with the algorithm it's configured with, not specifically that its the algorithms that the eIDAS spec specifies.

There's another test, `decryptAssertionShouldThrowIfWrongEncryptionAlgorithmUsed` that ensures that the handler throws if it encounters an assertion encrypted with an algorithm different to those listed in its validator.

src/main/java/uk/gov/ida/verifyserviceprovider/exceptions/MissingUnsignedAssertionsHandlerException.java
```diff
@@ -0,0 +1,7 @@
+package uk.gov.ida.verifyserviceprovider.exceptions;
+
+public class MissingUnsignedAssertionsHandlerException extends RuntimeException {
+    public MissingUnsignedAssertionsHandlerException() {
+        super("UnsignedAssertionsHandler missing: UnsignedAssertinosHandler can not be null in eIDAS enabled VSP");
```
galund: There's a typo here.

But I feel like this exception should not exist: surely we have a standard exception we can throw for bad configuration. This should not happen IRL, right? And I don't necessarily think it merits a test. either, surely we don't test that all of our code that expects a dependency to be injected haven't had `null` injected by mistake?

Wynndow: I had a look in the currently defined exceptions and there wasn't anything suitable, hence this. And I agree about not testing all our code that expects a dependency, but this is specifically about a a dependency that may or may not be null depending on a particular scenario.

PR #220, opened 2019-10-21
EID-1734 Add non-matching eIDAS CI manifest

Wynndow:
This PR may not actually be useful. See [this PR](https://github.com/alphagov/verify-terraform/pull/1084) on the release pipeline that consumes this manifest for detail.

We have three ways the VSP can be started:
* Matching
* Non-matching
* Non-matching with eIDAS

This new manifest will allow us to start up a non-matching eIDAS VSP in
the 'dev' environment.  We want this so we can test recent changes to
the VSP which allow it to accept unsigned assertions from eIDAS
countries.


Wynndow:
Have decided this approach is unnecessary, so closing.

PR #226, opened 2020-02-13
HUB-481: Improve error handling in Hub metadata config

CharlesIC:
- Upgrade Gradle to latest version

- Improve how Hub metadata configuration is parsed. The logic first tries to read overrides provided directly and then tries to read values from the environment enum if overrides have not been provided. This change introduces more checks and returns better error messages when neither the environment nor the override has been provided.
    
- It's now also possible to have an empty truststore section in the config template if the environment is provided. Previously the app would fail to initialise even if valid config was provided with environment set but empty truststore section.


src/main/java/uk/gov/ida/verifyserviceprovider/configuration/HubMetadataConfiguration.java
```diff
@@ -108,4 +110,22 @@ private KeyStore validateTruststore(KeyStore trustStore) {
         }
         return trustStore;
     }
+
+    private KeyStore tryGetTrustStore(TrustStoreConfiguration trustStoreConfiguration) {
+        try {
+            return trustStoreConfiguration.getTrustStore();
+        } catch (Exception ignored) {
```
Wynndow: What exception are you expecting to catch here, and why is it ignored? Also, seems odd that null is being returned.

Wynndow: Actually, the null thing isn't that odd, as it's being returned to `map()` which gives you an empty optional.

CharlesIC: It depends on the specific implementation - for instance for the `FileBackedTrustStoreConfiguration` it will attempt to load filestore even if the path is null - it doesn't do any checks. Those implementations live in saml-libs and are used in other places. I guess a "proper" fix would be to add check to the classes in saml-libs instead

CharlesIC: For instance here you'd get a NullPointerException
https://github.com/alphagov/verify-saml-libs/blob/c97880d1becb816d5b5a0d41a62ddf0da208e298/saml-lib/src/main/java/uk/gov/ida/saml/metadata/FileBackedTrustStoreConfiguration.java#L21

CharlesIC: OK actually the only case we want to ignore is NPE so I've changed this to only ignore that - this way if a value is provided but it's invalid it will still show up in the log and prevent the app from starting 👍

PR #230, opened 2020-08-17
Test errors for incorrectly crafted translation requests

instantiator:
* Adds a number of additional tests to `SuccessMatchAcceptanceTest` for errors that `translate-response` can generate.
* Adds some assertions to `AuthnRequestAcceptanceTest` to illustrate that the `requestId` is always generated.

### DIFF
--- a/src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/SuccessMatchAcceptanceTest.java
+++ b/src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/SuccessMatchAcceptanceTest.java
@@ -180,7 +180,7 @@ public class SuccessMatchAcceptanceTest {
     public void shouldReturnAnErrorWhenNullRequestIdIsProvided() {
         RequestResponseBody requestResponseBody = generateRequestService.generateAuthnRequest(application.getLocalPort());
 
-        Map<String, String> translateResponseRequestData = new HashMap<>(); // not a lot of map implementations allow nulls
+        Map<String, String> translateResponseRequestData = new HashMap<>();
         translateResponseRequestData.put(
             "samlResponse", complianceTool.createResponseFor(requestResponseBody.getSamlRequest(), BASIC_SUCCESSFUL_MATCH_WITH_LOA2_ID));
         translateResponseRequestData.put("requestId", null);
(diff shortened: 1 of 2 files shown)


src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/SuccessMatchAcceptanceTest.java
```diff
@@ -170,4 +175,147 @@ public void shouldReturnAnErrorWhenInvalidEntityIdProvided() {
         assertThat(errorMessage.getCode()).isEqualTo(BAD_REQUEST.getStatusCode());
         assertThat(errorMessage.getMessage()).isEqualTo("Provided entityId: invalidEntityId is not listed in config");
     }
+
+    @Test
+    public void shouldReturnAnErrorWhenNullRequestIdIsProvided() {
+        RequestResponseBody requestResponseBody = generateRequestService.generateAuthnRequest(application.getLocalPort());
+
+        Map<String, String> translateResponseRequestData = new HashMap<>(); // not a lot of map implementations allow nulls
```
JohnWattsGDS, line 183: Is this comment saying HashMap does allow nulls?  Is that a null value or a null key or both?  I'd understand if it was some obscure `ConcurrentFooBarRoboMap` or something as then the comment would imply the choice of Map implementation was because it accepted nulls but as `HashMap` is the go-to general purpose map, I don't think this comment is adding anything of value and actually detracts from the readability of the code (I literally can't concentrate on anything else because I'm wondering what it's telling me that I don't understand).  And this is coming from someone who isn't dogmatic about "no comments".  What am I missing?

instantiator: I was explaining why I hadn't used the same pattern that appears in the all the existing tests - because `HashMap` seems to be just about the only implementation of Map that allows a null value.

instantiator: More than happy to remove.

JohnWattsGDS: I'd committed the sin of not reviewing the entire file.  However, if it had said `// ImmutableMap does not allow null values` then I'd have probably known what was going on without reading the whole file.  😉

instantiator: Updated - no worries. Comment wasn't really necessary.

PR #232, opened 2020-09-14
HUB-697: Update truststores and fix connector node metadata URL

Wynndow:
This PR adds the new G3 CAs from the [Verify pki certs repo](https://github.com/alphagov/verify-pki-certs) to the production and test truststores. It also adds the new generation of [dev-pki certificates](https://github.com/alphagov/verify-dev-pki) to the test truststores.

To review, the password for the truststores is `bj76LWZ+F5L1Biq4EZB+Ta7MUY4EQMgmZmqAHh` (don't worry, it's already open sourced). Having said that, you don't actually need the password to view the contents of the truststores. Shrug.

It also updates the URL for the connector-node metadata. This has now been decided as we have the qwac cert for the connector-node itself. This means we can drop the other possibilities. It would be good for @idrop to review and confirm this is ok.


idrop (dismissed):
👍 Ok with the VMC generated connector node urls.

idrop (approved):

instantiator (approved):

PR #238, opened 2021-01-19
HUB-774 Update Dockerfile to use GHCR

j-maynard:
Update the Dockerfile to use images from GitHub container registry rather than having them come from DockerHub.

### DIFF
--- a/Dockerfile.internal
+++ b/Dockerfile.internal
@@ -1,4 +1,4 @@
-FROM gradle:6.7.0-jdk11 as build
+FROM ghcr.io/alphagov/verify/gradle:gradle-jdk11 as build
 ARG VERIFY_USE_PUBLIC_BINARIES=false
 WORKDIR /verify-service-provider
 USER root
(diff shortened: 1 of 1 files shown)


instantiator (approved):

idrop (approved):

Dockerfile.internal
```diff
@@ -1,4 +1,4 @@
-FROM gradle:6.7.0-jdk11 as build
+FROM ghcr.io/alphagov/verify/gradle:gradle-jdk11 as build
```
galund, line 1: Shouldn't we still be specifying the gradle version here? If we want to upgrade, we need a managed way to do that, without breaking everything at once?

CharlesIC: We'd have to change the way we vendor libraries to include a version tag in the vendored images. At the moment, we don't even specify which version of gradle to pull from dockerhub, we just use the tag `jdk11`

https://github.com/alphagov/verify-terraform/blob/e1029473ed20816ad20faa34da4b2f0ed4fe825f/environments/tools/platform/files/pipelines/vendor.yml#L268-L281

PR #239, opened 2021-02-10
EID-2087 Remove eIDAS code and config from the VSP

idrop:
Remove processing of assertions received from EU Member States.

The configuration parameter `europeanIdentity`, if present, and child elements must be removed by services in order for the VSP to start.


RELEASE_NOTES.md
```diff
@@ -1,9 +1,9 @@
 Release notes
 =============
 
-### Next
-* Fixes an issue with trust anchors for eIDAS metadata resolvers not being updated when they changed at source. The VSP
-  will now update them without requiring a restart.
```
galund, line 6: Arguably this is a backwards-incompatible change so we should be giving it a new major version number? (Otherwise we could perhaps just make it ignore the `europeanIdentity` config?) Not that this really matters much. Bravo :clap:

RELEASE_NOTES.md
```diff
@@ -1,9 +1,9 @@
 Release notes
 =============
 
-### Next
```
galund, line 4: Oh, and you should leave the existing release notes in place, and the heading "next". Then you do a separate PR to release the lot, if I remember correctly.

galund (changes requested):

RELEASE_NOTES.md
Wynndow: Wot George said ^

RELEASE_NOTES.md
Wynndow: I agree, should be v3.0.0. But you set that when doing the release.

build.gradle
```diff
@@ -31,7 +31,7 @@ repositories {
 }
 
 project.ext {
-    version_number = '2.2.0'
+    version_number = '2.3.0'
```
Wynndow, line 34: I think the release scripts set this for you.

build.gradle
```diff
@@ -158,8 +158,6 @@ distributions {
                 [
                     'configuration/vsp-with-matching.yml',
                     'configuration/vsp-with-matching-custom.yml',
-                    'configuration/vsp-no-eidas.yml',
```
Wynndow, line 161: ~Should the vsp without eidas config be kept? Maybe it could be renamed though?~
Looks like it's exactly the same as the `verify-service-provider.yml` config except with the `europeanIdentity` key still in so probably fine.

src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/NonMatchingAcceptanceTest.java
```diff
@@ -260,90 +252,6 @@ public void shouldThrowExceptionIfLoAReturnedByIdpIsTooLow () {
...
-
-        assertThat(response.getStatus()).isEqualTo(OK.getStatusCode());
-
-        JSONObject jsonResponse = new JSONObject(response.readEntity(String.class));
-        assertThat(jsonResponse.getString("scenario")).isEqualTo(IDENTITY_VERIFIED.name());
-        assertThat(jsonResponse.getString("levelOfAssurance")).isEqualTo(LEVEL_1.name());
-    }
-
-    @Test
-    public void shouldProcessIdpResponseCorrectlyWhenEuropeanIdentityConfigAbsent() {
```
Wynndow, line 320: This test seems to be exercising the non eIDAS app - should it be kept?

src/main/java/uk/gov/ida/verifyserviceprovider/factories/VerifyServiceProviderFactory.java
```diff
@@ -136,53 +114,6 @@ private TranslateSamlResponseResource getTranslateMatchingSamlResponseResource()
         return new TranslateSamlResponseResource(matchingResponseService, entityIdService);
     }
 
-    private TranslateSamlResponseResource getEidasEnabledTranslateNonMatchingSamlResponseResource() {
```
Wynndow, line 139: 😍

src/test/java/unit/uk/gov/ida/verifyserviceprovider/services/ClassifyingAssertionTranslatorTest.java
```diff
@@ -32,83 +28,25 @@
...
 
-    @Mock
-    private EidasAssertionTranslator eidasAssertionService;
-
-    @Mock
-    private EidasUnsignedAssertionTranslator eidasUnsignedAssertionTranslator;
-
     @Before
     public void setUp() {
         initMocks(this);
```
Wynndow, line 33: Is `initMocks()` required? If not the whole `setUp` method could go.

src/main/java/uk/gov/ida/verifyserviceprovider/services/ClassifyingAssertionTranslator.java
```diff
@@ -9,38 +9,17 @@
 public class ClassifyingAssertionTranslator implements AssertionTranslator {
 
     private final VerifyAssertionTranslator verifyAssertionTranslator;
-    private final EidasAssertionTranslator eidasAssertionTranslator;
-    private final EidasUnsignedAssertionTranslator eidasUnsignedAssertionTranslator;
 
 
     public ClassifyingAssertionTranslator(
```
Wynndow, line 14: This class just seems to be a wrapper around the VerifyAssertionTranslator class. Could we cut out the middleman and drop this class completely?

RELEASE_NOTES.md
idrop: Thank, I've reverted so the release notes are back in ### Next

build.gradle
idrop: Now addressed, thanks

src/acceptance-test/java/uk/gov/ida/verifyserviceprovider/NonMatchingAcceptanceTest.java
idrop: I've restored, thanks!

src/test/java/unit/uk/gov/ida/verifyserviceprovider/services/ClassifyingAssertionTranslatorTest.java
idrop: The tests fail if initMocks isn't there

src/main/java/uk/gov/ida/verifyserviceprovider/services/ClassifyingAssertionTranslator.java
idrop: Yep, removed this class with no affect

Wynndow (approved):

RELEASE_NOTES.md
```diff
@@ -2,8 +2,7 @@ Release notes
 =============
 
 ### Next
-* Fixes an issue with trust anchors for eIDAS metadata resolvers not being updated when they changed at source. The VSP
-  will now update them without requiring a restart.
+* Removed processing of assertions received from EU Member States. The configuration parameter `europeanIdentity`, if present, and child elements must be removed in order for the VSP to start.
```
galund, line 5: I think I'm confused that we're still imposing this backwards-incompatible change on services - have they all confirmed this is ok? If not can't we split that off?

idrop: Unlike the MSA, we can remove all config from the VSP, and it won't affect the processing of Verify SAML assertions. The presence of eIDAS config in the MSA changes the payload sent to the local matching service.

I think it's reasonable to tell VSP users they have to remove all eIDAS config along with the upgrade.

galund: Aha, I was confused which PR I was reading :-D

galund: (To be honest I don't really see why we can't just _ignore_ the `europeanIdentity` config rather than insist they make a change, but :shrug: :-) )

galund (approved):

PR #248, opened 2022-01-05
BAU Upgrade to Dropwizard 2

JohnWattsGDS:
Update Dropwizard and the shared test utils.
Change some expected error messages in tests.
Add some `throws Exception`s to test methods
Explicitly add an ObjectMapper that doesn't like unknown properties to a test that needed it.

### DIFF
--- a/src/main/java/uk/gov/ida/verifyserviceprovider/configuration/VerifyServiceProviderConfiguration.java
+++ b/src/main/java/uk/gov/ida/verifyserviceprovider/configuration/VerifyServiceProviderConfiguration.java
@@ -3,11 +3,9 @@ package uk.gov.ida.verifyserviceprovider.configuration;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import io.dropwizard.Configuration;
-import org.hibernate.validator.valuehandling.UnwrapValidatedValue;
 import org.joda.time.Duration;
 import uk.gov.ida.saml.metadata.MetadataResolverConfiguration;
 import uk.gov.ida.verifyserviceprovider.exceptions.NoHashingEntityIdIsProvidedError;
-
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
 import javax.validation.constraints.Size;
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/ida/verifyserviceprovider/configuration/VerifyServiceProviderConfiguration.java
```diff
@@ -39,7 +37,7 @@ public VerifyServiceProviderConfiguration(
             @JsonProperty("samlSigningKey") @NotNull @Valid @JsonDeserialize(using = PrivateKeyDeserializer.class) PrivateKey samlSigningKey,
             @JsonProperty("samlPrimaryEncryptionKey") @NotNull @Valid @JsonDeserialize(using = PrivateKeyDeserializer.class) PrivateKey samlPrimaryEncryptionKey,
             @JsonProperty("samlSecondaryEncryptionKey") @Valid @JsonDeserialize(using = PrivateKeyDeserializer.class) PrivateKey samlSecondaryEncryptionKey,
-            @JsonProperty("msaMetadata") @NotNull @UnwrapValidatedValue @Valid Optional<MsaMetadataConfiguration> msaMetadata,
+            @JsonProperty("msaMetadata") @NotNull Optional<@NotNull @Valid MsaMetadataConfiguration> msaMetadata,
```
Wynndow, line 40: My head is struggling to understand what's happening here?

Wynndow:
I was wondering why you didn't have to do any of the injection jiggery pokery like with the MSA and the hub. But then realised that the VSP doesn't use injection. Nice.

Might be worth adding a line to `RELEASE_NOTES.md` before merging.

src/main/java/uk/gov/ida/verifyserviceprovider/configuration/VerifyServiceProviderConfiguration.java
JohnWattsGDS: Me too.  Before @UnwrapValidatedValue extracted a thing from a wrapper (like Optional) and then did the validation on the extracted value.  That annotation disappeared.  As I understand it, you can now put your annotations next to the wrapped type and it all just works.  At first I took out the original @NotNull but figured the optional value should not be null as well as the wrapped value so we're making sure there's a value for that and then validating the contents.

CharlesIC (approved):

Wynndow (approved):

